PR #4, opened 2021-08-19
Sergey/rest api support

sergeyberezansky:


pkg/wekafs/controllerserver.go
```diff
@@ -191,81 +212,128 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+	return &csi.ControllerExpandVolumeResponse{}, err
 }
 
 func (cs *controllerServer) ControllerExpandVolume(ctx context.Context, req *csi.ControllerExpandVolumeRequest) (*csi.ControllerExpandVolumeResponse, error) {
 
 	if len(req.GetVolumeId()) == 0 {
 		return nil, status.Errorf(codes.InvalidArgument, "Volume ID not specified")
 	}
-	volume, err := NewVolume(req.GetVolumeId())
+	client, err := cs.api.GetClientFromSecrets(req.Secrets)
```
sergeyberezansky, line 283: need to check if this is not causing issues with backward compatibility, e.g. 
delete still works for volumes without API

sergeyberezansky: Apparently not, since in this case we `get client, err := nil, nil`

pkg/wekafs/dirvolume.go
```diff
@@ -0,0 +1,388 @@
...
+		glog.Errorf("Volume %s is unusable: path %s is a not a directory", v.GetId(), v.Filesystem)
+		return false, status.Error(codes.Internal, err.Error())
+	}
+	glog.Infof("Volume %s exists and accessible via %s", v.id, v.getFullPath(mountPoint))
+	return true, nil
+}
+
+func (v DirVolume) Create(mountPath string, enforceCapacity bool, capacity int64) error {
+	volPath := v.getFullPath(mountPath)
+	if err := os.MkdirAll(volPath, 0750); err != nil {
```
sergeyberezansky, line 346: We have been asked few times around the default permissions we give to CSI volumes
Should we make it configurable in SC?

tests/csi-sanity/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM golang
 #ENV test=test
-RUN go get github.com/kubernetes-csi/csi-test/cmd/csi-sanity
+RUN GO_ENABLED=0 GOOS=linux GOARCH=amd64 go get -u -t github.com/kubernetes-csi/csi-test/cmd/csi-sanity || true
```
sergeyberezansky, line 3: @tigrawap, I've recalled the problem I have had with csi-sanity test in container.
Apparently, the `go get github.com/kubernetes-csi/csi-test/cmd/csi-sanity` command constantly fails due to error that the package contains only tests

I could only find some info on this here: https://github.com/golang/go/issues/29268#
I've tried to somehow work around it - but it fails (not only returns invalid retcode, but actually doesn't install the package)
Ideas are welcome!

pkg/wekafs/apiclient/apiclient.go
```diff
@@ -0,0 +1,781 @@
...
+	}
+}
+
+// handleNetworkErrors checks if the error returned by endpoint is a network error (transient by definition)
+func (a *ApiClient) handleNetworkErrors(err error) error {
+	if err == nil {
+		return nil
+	}
+	if netError, ok := err.(net.Error); ok && netError.Timeout() {
+		println("Timeout")
```
tigrawap, line 444: log and not print

pkg/wekafs/apiclient/apiclient.go
```diff
@@ -0,0 +1,781 @@
...
+		return err
+	} else {
+		switch t := err.(type) {
+		case *net.OpError:
+			if t.Op == "dial" {
+				println("Unknown host")
```
tigrawap, line 450: return error somewhere?

pkg/wekafs/dirvolume.go
```diff
@@ -0,0 +1,388 @@
...
+	maxCapacity := int64(stat.Bavail * uint64(stat.Bsize))
+	glog.V(4).Infoln("Max capacity available for a volume:", maxCapacity)
+	return maxCapacity, nil
+}
+
+func (v DirVolume) GetType() VolumeType {
+	return VolumeTypeDirV1
+}
+
+func (v DirVolume) GetCapacity(mountPath string) (int64, error) {
```
tigrawap, line 47: - handle error from getSizeFromQuota. On 500 and 404 fall back to xattr (no need for df)
- if no client - use xattr
- why do we need df? can't come up with scenario

pkg/wekafs/dirvolume.go
```diff
@@ -0,0 +1,388 @@
...
+	if err == nil {
+		glog.V(3).Infoln("Current capacity of volume", v.GetId(), "is", size, "obtained via fs stat")
+		return int64(size), nil
+	}
+	return 0, errors.New("could not fetch volume capacity")
+}
+
+func (v DirVolume) UpdateCapacity(mountPath string, enforceCapacity *bool, capacityLimit int64) error {
+	glog.V(3).Infoln("Updating capacity of the volume", v.GetId(), "to", capacityLimit)
+	f := func() error { return v.updateCapacityQuota(mountPath, enforceCapacity, capacityLimit) }
```
tigrawap, line 73: If received 404 - try to fallback to xattrs

pkg/wekafs/dirvolume.go
```diff
@@ -0,0 +1,388 @@
...
+	fs, err := v.getFilesystemObj()
+	if err != nil {
+		glog.Errorln("Failed to fetch filesystem for volume", v.GetId())
+		return err
+	}
+
+	q, err := v.apiClient.GetQuotaByFileSystemAndInode(fs, inodeId)
+	// check if the quota already exists. If not - create it and exit
+	if err != nil {
+		if err != apiclient.ObjectNotFoundError {
```
tigrawap, line 108: you could use this for falling back to xattr if quota not found, i.e legacy volue

pkg/wekafs/dirvolume.go
```diff
@@ -0,0 +1,388 @@
...
+		if capacity, err := strconv.ParseInt(string(capacityString), 10, 64); err == nil {
+			return uint64(capacity), nil
+		}
+		return 0, ErrBadXattrOnVolume
+	}
+	return 0, ErrNoXattrOnVolume
+}
+
+//getSizeFromDf used for GetCapacity, since we don't have secrets to check API
+func (v DirVolume) getSizeFromDf(mountPath string) (uint64, error) {
```
tigrawap, line 295: looks like not needed

pkg/wekafs/dirvolume.go
tigrawap: if it's doable - why not

pkg/wekafs/gc.go
```diff
@@ -1,5 +1,6 @@
 package wekafs
 
+// TODO: fix emplementation of VolumeGC
```
tigrawap, line 3: double check if anything actually left

sergeyberezansky: Removing

pkg/wekafs/dirvolume.go
sergeyberezansky: It is too complex. MEanwhile TODO

pkg/wekafs/dirvolume.go
sergeyberezansky: Did this to support volume.GetCapacity() for ControllerServer.GetCapacity() but it can't be done anyway.

pkg/wekafs/dirvolume.go
sergeyberezansky: yes. did it on updateCapacity flow but not in GetCapacity

dontbreakit:
here

PR #7, opened 2022-02-15
Sergey/auth mounts

sergeyberezansky:
Support for authenticated mounts (implicit for non-root organization, explicit with auth-required setting)
Requires Weka software version 3.14 and up


README.md
```diff
@@ -1,5 +1,5 @@
 # CSI WekaFS Driver
-![Version: 0.7.3](https://img.shields.io/badge/Version-0.7.3-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.7.3](https://img.shields.io/badge/AppVersion-v0.7.3-informational?style=flat-square)
+![Version: 0.7.4-beta](https://img.shields.io/badge/Version-0.7.4--beta-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.7.4-beta](https://img.shields.io/badge/AppVersion-v0.7.4--beta-informational?style=flat-square)
```
tigrawap, line 2: why beta in code? 
beta is manually created release...code should contain only actual versions

deploy/helm/csi-wekafsplugin/Chart.yaml
```diff
@@ -2,9 +2,9 @@ apiVersion: v2
...
 description: Helm chart for Deployment of WekaIO Container Storage Interface (CSI) plugin for WekaFS - the world fastest filesystem
 sources:
-  - https://github.com/weka/csi-wekafs/tree/v0.7.3/deploy/helm/csi-wekafsplugin
+  - https://github.com/weka/csi-wekafs/tree/v0.7.4-beta/deploy/helm/csi-wekafsplugin
 home: https://github.com/weka/csi-wekafs
 icon: https://weka.github.io/csi-wekafs/logo.png
 type: application
-version: 0.7.3
-appVersion: v0.7.3
+version: 0.7.4-beta
```
tigrawap, line 9: same here

pkg/wekafs/apiclient/apiclient.go
```diff
@@ -94,11 +97,13 @@ func (cm *WekaCompatibilityMap) fillIn(versionStr string) {
 	f, err := version.NewVersion(MinimumSupportedWekaVersions.FilesystemAsVolume)
 	q, err := version.NewVersion(MinimumSupportedWekaVersions.QuotaDirectoryAsVolume)
 	n, err := version.NewVersion(MinimumSupportedWekaVersions.QuotaOnNonEmptyDirs)
+	a, err := version.NewVersion(MinimumSupportedWekaVersions.AuthenticatedMount)
```
tigrawap, line 100: better names please

tigrawap: also, errors are not handled

pkg/wekafs/apiclient/filesystem.go
```diff
@@ -135,6 +141,20 @@ func (a *ApiClient) DeleteFileSystem(r *FileSystemDeleteRequest) error {
 	return nil
 }
 
+func (a *ApiClient) GetFileSystemMountToken(r *FileSystemMountTokenRequest, token *FileSystemMountToken) error {
+	f := a.Log(5, "Getting a mount token for filesystem", r.Uid)
```
tigrawap, line 145: why not constants for log levels?

tigrawap: also, what's a poing of deferring this logline? 
also not sure..how comes `Log` function returns a function? Kinda does not make sense

pkg/wekafs/mounter.go
```diff
@@ -79,14 +82,55 @@ func (m *wekaMount) doUnmount() error {
...
+				glog.V(3).Infof("API client not supports authenticated mounts, skipping")
+			} else {
+				glog.V(3).Infof("Requesting mount token for filesystem %s via API", m.fsRequest.fs)
+				filesystem, err := apiClient.GetFileSystemByName(m.fsRequest.fs)
+				if err != nil {
+					return err
+				}
+				if filesystem != nil {
+					req := &apiclient.FileSystemMountTokenRequest{Uid: filesystem.Uid}
+					token := &apiclient.FileSystemMountToken{}
```
tigrawap, line 105: why not just create and return in function with cleaner signature, instead of passing references (and better to avoid use of pointers, when not really required. Does not seem really required to me)

pkg/wekafs/mounter.go
```diff
@@ -139,11 +183,11 @@ func (m *wekaMounter) initFsMountObject(fs fsRequest) {
 
 type UnmountFunc func()
 
-func (m *wekaMounter) mountParams(fs string, xattr bool) (string, error, UnmountFunc) {
-	request := fsRequest{fs, xattr}
+func (m *wekaMounter) mountParams(fs string, xattr bool, apiClient *apiclient.ApiClient) (string, error, UnmountFunc) {
```
tigrawap, line 186: nitpick, but if apiClient can't be part of wekaMounter itself, then its better to have db references, API clients and so on as earlier arguments, and not last one

pkg/wekafs/mounter.go
sergeyberezansky: Made for retain signature similarity with other ApiClient functions. Unless I don't get you right.

Pointer required for cases where this is not supported, the generic mount function will either get a valid token or nil

pkg/wekafs/mounter.go
sergeyberezansky: since this param is "optional", put it last

README.md
sergeyberezansky: Will drop this commit

pkg/wekafs/mounter.go
tigrawap: but this is not generic function
it's `apiClient` function that returns struct that also part of apiclient
it would make more sense just to create and return token
passing pointer makes sense for generalistion when it's not specific struct, but interfaces. not a case here

pkg/wekafs/mounter.go
tigrawap: but it is not optional
`mounter.incRef(apiClient)` you use it here directly, there is no logic for null

tigrawap: ok, found that it's optional much much deeper
i feel that additional incapsulation is missing here
too much of not clear passing of pointers.. but ok, another time

tigrawap (approved):

PR #30, opened 2023-02-27
ci: add csi-sanity action to PRs

dontbreakit:


tests/csi-sanity/ga-docker-compose1.yaml
```diff
@@ -0,0 +1,31 @@
+version: '3'
```
sergeyberezansky, line 1: suggest to rename docker-compose-nosnapshotcaps.yaml
Why do you have ga- prefix?

dontbreakit: will rename

.github/workflows/sanity.yaml
```diff
@@ -0,0 +1,90 @@
...
+    needs: build
+    runs-on: self-hosted
+    steps:
+      - uses: actions/checkout@v3
+      - run: docker-compose -f tests/csi-sanity/docker-compose-nosnapshotcaps.yaml up $COMPOSE_DEFAULTS
+        env:
+          SANITY_FUNCTION: legacy_sanity
+            
+  directory_volume_no_snapshots:
+    if: success() || failure() # always() can't be canceled
```
dontbreakit, line 43: technically the same as `always()` but can be canceled

sergeyberezansky (approved):

PR #34, opened 2023-03-13
ci: add release action

dontbreakit:
> feature to dev = squash > release if any
   dev to master = merge > release if any

- workflow triggering
image

- example of artifacthub
image

- example of GitHub releases with autogenerated changelog
image


deploy/helm/csi-wekafsplugin/Chart.yaml
```diff
@@ -5,14 +5,13 @@ maintainers:
   - name: WekaIO, Inc.
     email: csi@weka.io
     url: https://weka.io
-sources:
```
sergeyberezansky, line 8: Why removing from here?

dontbreakit: the field is optional which should lead to the project, not to a release. And the Homepage link is already pointing to the project/repo (with all releases) @sergeyberezansky 

image

dontbreakit: updated it back to point to a tag

PR #40, opened 2023-04-02
fix: Allow ExpandVolume when current capacity is not set

sergeyberezansky:


sergeyberezansky:
Cannot be done since setting quota is not supported on non-empty volume

PR #494, opened 2025-03-18
feat(CSI-356): avoid failback to xattr upon quota set error

sergeyberezansky:
### TL;DR

Simplified volume capacity update logic by removing the fallback mechanism and improving logging.

### What changed?

- Removed the fallback mechanism that would try to use xattr if setting quota failed
- Added a `capacityEntity` variable to track which method is being used (quota or xattr)
- Improved logging by including the capacity entity in success and error messages
- Enhanced warning message for non-default organization to suggest upgrading to the latest Weka software
- Streamlined error handling logic to be more consistent

### How to test?

1. Test volume capacity updates in different scenarios:
   - With a regular Weka cluster that supports quota
   - In in DEV mode
   - With a volume that has no API client bound
   - With a cluster that doesn't support quota directory as volume
   - With a non-default organization
   - With a snapshot-backed volume

2. Verify that appropriate logs are generated with the correct capacity entity. xattr should be used only when using quota is not supported, not when it fails with error.

### Why make this change?

The previous implementation had unnecessary complexity with a fallback mechanism that wasn't being used effectively. As a result, upon failure to set up quota a volume would be created using xattr failback mode, possibly concealing the fact that the quota is not actually set. This change simplifies the code flow, makes the behavior more predictable, and improves logging to make it clearer which capacity update method is being used, making troubleshooting easier.

### DIFF
--- a/pkg/wekafs/volume.go
+++ b/pkg/wekafs/volume.go
@@ -541,40 +541,33 @@ func (v *Volume) UpdateCapacity(ctx context.Context, enforceCapacity *bool, capa
 
 	// update capacity of the volume by updating quota object on its root directory (or XATTR)
 	logger.Info().Int64("desired_capacity", capacityLimit).Msg("Updating volume capacity")
-	var fallback = true // whether we should try to use xAttr fallback or not if setting quota was attempted and failed. note that in certain cases a fallbackFunc will be used right away:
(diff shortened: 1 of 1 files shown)


sergeyberezansky:
### Merge activity

* **Mar 18, 12:51 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/weka/csi-wekafs/494).

PR #495, opened 2025-03-18
fix(CSI-357): server default mount options take precedence over custom ones

sergeyberezansky:
### TL;DR

Fixed mount options precedence and enhanced mutually exclusive options handling.

### What changed?

1. Reversed the order of mount options merging in `Volume.MountUnderlyingFS()` to prioritize volume-specific options over server default options
2. Added a new mutually exclusive mount option set for caching-related options: `coherent`, `forcedirect`, `readcache`, and `writecache`.
3. Enhanced the test case for `MountOptions_Merge` to verify the behavior of the new mutually exclusive options.

### How to test?

1. Run the updated unit tests: `go test ./pkg/wekafs -run TestMountOptions_Merge`
2. Verify that when merging mount options, the mutually exclusive options are properly handled (e.g., `writecache` is removed when `forcedirect` is added).
3. Test mounting a volume with conflicting options to ensure the server defaults take precedence.

### Why make this change?

This change ensures proper handling of mount options with the correct precedence. Server default options should be applied first, then potentially overridden by volume-specific options. Additionally, the new mutually exclusive set prevents incompatible caching options from being used together, which could lead to unexpected behavior or errors during mounting.

### DIFF
--- a/pkg/wekafs/mountoptions_test.go
+++ b/pkg/wekafs/mountoptions_test.go
@@ -23,11 +23,12 @@ func TestMountOptions_RemoveOption(t *testing.T) {
 }
 
 func TestMountOptions_Merge(t *testing.T) {
-	opts1 := NewMountOptions([]string{"ro"})
-	opts2 := NewMountOptions([]string{"rw"})
+	opts1 := NewMountOptions([]string{"ro", "writecache"})
+	opts2 := NewMountOptions([]string{"rw", "forcedirect"})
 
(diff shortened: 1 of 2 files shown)


sergeyberezansky:
### Merge activity

* **Mar 18, 12:51 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/weka/csi-wekafs/495).

PR #550, opened 2025-07-23
fix(CSI-375): deletion of volume may be stuck if its contents were already trashed

sergeyberezansky:
### TL;DR

Improved error handling in the volume garbage collection process to handle edge cases.

### What changed?

- Added proper error handling when creating the garbage collection directory, returning an error only if it's not an "already exists" error
- Added a check to verify if the volume path exists before attempting to move it to trash
- If the volume path doesn't exist (possibly already moved), the function now logs a debug message and returns successfully instead of failing

### How to test?

1. Test garbage collection with a volume that doesn't exist
2. Test garbage collection when the trash directory already exists
3. Test normal garbage collection flow with existing volumes

### Why make this change?
In situations where garbage collection was invoked but failed to start (e.g. due to WEKA cluster API inavailability or networking issue), volumes could remain in "Deleting" state forever in Kubernetes.

This change makes the garbage collection process more robust by handling edge cases that could cause unnecessary errors:
- When the trash directory already exists, we shouldn't treat it as an error
- When a volume's contents are already missing, we should log and continue rather than failing the operation

### DIFF
--- a/pkg/wekafs/gc.go
+++ b/pkg/wekafs/gc.go
@@ -65,13 +65,18 @@ func (gc *innerPathVolGc) moveVolumeToTrash(ctx context.Context, volume *Volume)
 	}
 	volumeTrashLoc := filepath.Join(path, garbagePath)
 	if err := os.MkdirAll(volumeTrashLoc, DefaultVolumePermissions); err != nil {
-		logger.Error().Str("garbage_collection_path", volumeTrashLoc).Err(err).Msg("Failed to create garbage collector directory")
-	} else {
-		logger.Debug().Str("garbage_collection_path", volumeTrashLoc).Msg("Successfully created garbage collection directory")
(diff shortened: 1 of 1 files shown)


sergeyberezansky:
* **#550** <a href="https://app.graphite.dev/github/pr/weka/csi-wekafs/550?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/weka/csi-wekafs/550?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

sergeyberezansky:
### Merge activity

* **Jul 27, 11:57 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/weka/csi-wekafs/550).

PR #591, opened 2025-08-06
fix: readiness probe might fail under load due to too small max concurrency of GRPC server

sergeyberezansky:
### TL;DR

Updated the description of `maxConcurrentRequests` parameter and implemented dynamic calculation of gRPC server's `MaxConcurrentStreams`.

### What changed?

- Clarified the description of `maxConcurrentRequests` in the Helm chart values file to indicate it applies "for each sidecar" rather than "global"
- Added logic to dynamically calculate the appropriate value for `MaxConcurrentStreams` in the gRPC server options:
  - Sums all maximum concurrency values per operation
  - Doubles the result to provide extra capacity for liveness probes and other operations
  - Sets a minimum value of 256 if the calculated value is lower

### How to test?

1. Deploy the CSI plugin with various concurrency settings
2. Verify that the gRPC server correctly handles the expected number of concurrent requests
3. Check that under high load, the server properly manages request concurrency according to the configured limits

### Why make this change?

This change improves the server's ability to handle concurrent requests by dynamically setting the `MaxConcurrentStreams` parameter based on the actual concurrency configuration. The previous implementation didn't properly account for the total number of potential concurrent operations, which could lead to connection issues under high load. The updated parameter description also provides clearer guidance to users configuring the Helm chart.

### DIFF
--- a/charts/csi-wekafsplugin/values.yaml
+++ b/charts/csi-wekafsplugin/values.yaml
@@ -50,7 +50,7 @@ machineConfigLabels:
 controller:
   # -- Controller number of replicas
   replicas: 2
-  # -- Maximum concurrent requests from sidecars (global)
+  # -- Maximum concurrent requests from sidecars (for each sidecar)
   maxConcurrentRequests: 25
   # -- maximum concurrent operations per operation type
   concurrency:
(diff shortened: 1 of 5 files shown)


sergeyberezansky:
### Merge activity

* **Aug 6, 6:06 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/weka/csi-wekafs/591).

PR #622, opened 2025-11-05
feat(CSI-316): support per-filesystem encryption with custom KMS settings

assafgi:
This feature extends the existing encryption support (CSI-315) by adding the ability to specify custom KMS credentials per filesystem, rather than using only the cluster-wide KMS configuration.

Key changes:
- Added KMS credential fields to Volume struct for per-filesystem encryption key management
- Extended ApiClient to support KMS credentials from secrets
- Added methods to enrich volumes with KMS credentials from API secrets
- Enhanced encryption parameter validation to support custom KMS settings
- Added support for reading KMS credentials (role_id, secret_id, namespace, key_identifier) from CSI secrets
- Improved isEncrypted() method to populate KMS params from existing filesystems
- Added masking for sensitive KMS fields in logs

New examples:
- Added dynamic_filesystem_encryption_key_in_secret example showing how to use per-filesystem encryption

This allows users to:
1. Use different KMS keys for different filesystems
2. Manage KMS credentials per StorageClass via Kubernetes secrets
3. Override cluster-wide KMS settings on a per-volume basis

Requirements:
- Weka cluster must support custom encryption settings
- KMS server must be configured to allow per-filesystem keys
- API secret must include KMS credentials in the expected format

### DIFF
--- a/pkg/wekafs/volume.go
+++ b/pkg/wekafs/volume.go
@@ -52,9 +52,14 @@ type Volume struct {
 	enforceCapacity       bool
 	initialFilesystemSize int64
 	mountOptions          MountOptions
-	encrypted             bool
+	encrypted             *bool // to support also encryption state fetched from actual filesystem when not set
 	manageEncryptionKeys  bool
 	encryptWithoutKms     bool
+
+	kmsVaultNamespace     string
+	kmsVaultKeyIdentifier string
+	kmsVaultRoleId        string
+	KmsVaultSecretId      string
 
 	srcVolume   *Volume
 	srcSnapshot *Snapshot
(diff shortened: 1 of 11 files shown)


pkg/wekafs/volume.go
```diff
@@ -1737,10 +1854,10 @@ func (v *Volume) ObtainRequestParams(ctx context.Context, params map[string]stri
 		if err != nil {
 			return errors.Join(err, errors.New("failed to parse 'encrypted' parameter"))
 		}
-		v.encrypted = encrypted
+		v.encrypted = &encrypted
 	}
```
sergeyberezansky: relates to comments on line 1870. maybe if &encrypted is false, just skip.

sergeyberezansky (changes requested):
@assafgi I think I missed a part where secret has encryption keys but storageClass defines encrypted=false.

Please look at my comments

sergeyberezansky (approved):

PR #623, opened 2025-11-06
chore: debug image

rugggger:
# Add Debug Tooling for CSI WekaFS Plugin

This PR adds comprehensive debugging capabilities to the CSI WekaFS plugin:

- Adds a `debug.Dockerfile` that includes Delve debugger and debug symbols
- Creates new Makefile targets:
  - `build-debug-binaries`: Builds binaries with debug symbols
  - `build-debug`: Creates a debug Docker image with Delve
  - `deploy-debug`: Complete workflow that:
    - Builds and pushes debug image to Quay.io
    - Updates the Kubernetes deployment to use the debug image
    - Configures Delve and removes liveness probes
    - Provides instructions for connecting debuggers

The debug image is tagged with a timestamp and pushed to `quay.io/weka.io/csi-wekafs-debug`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,11 +15,16 @@
 CMDS=wekafsplugin
 all: build
 
-.PHONY: build-% build clean
+.PHONY: build-% build clean build-debug deploy-debug
 
 # understand what is the version tag
 VERSION ?= $(shell cat charts/csi-wekafsplugin/Chart.yaml | grep appVersion | awk '{print $$2}' | tr -d '"')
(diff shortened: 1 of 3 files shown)


assafgi (approved):

sergeyberezansky (commented):
@rugggger I don't like the idea of creating a separate docker image.
The metrics server in 3.0 already incorporates debug mode set via env var.
I think that this is the correct way to go here as well.

Also, you try to build a metricsserver that does not exist in 2.8

sergeyberezansky (approved):

PR #631, opened 2025-11-28
fix: initialize API metrics once at startup (CSI-398)

assafgi:
### TL;DR

Refactored API metrics collection to use a global singleton pattern with proper initialization.

### What changed?

- Removed the `metrics` field from the `ApiClient` struct
- Replaced instance-based metrics with a global singleton `apiMetrics` variable
- Added an `InitApiMetrics()` function that's called during package initialization
- Removed unnecessary blank lines in initialization functions

### How to test?

1. Run the metrics server or the CSI plugin and verify that API metrics are still being collected and exposed properly
2. Check that the Prometheus endpoint shows the expected API request counters and durations
3. Verify that no errors occur during initialization or when making API requests

### Why make this change?

This change avoids a race condition where multiple API clients for different sets of credentials are agressively initalized in parallel.
This prevents potential issues with multiple metric registrations and ensures metrics are properly initialized before use.

### DIFF
--- a/cmd/metricsserver/main.go
+++ b/cmd/metricsserver/main.go
@@ -33,7 +33,6 @@ func init() {
 		file = short
 		return file + ":" + strconv.Itoa(line)
 	}
-
 }
 
 var (
--- a/cmd/wekafsplugin/main.go
+++ b/cmd/wekafsplugin/main.go
@@ -49,7 +49,6 @@ func init() {
 		file = short
 		return file + ":" + strconv.Itoa(line)
 	}
-
 }
 
 var (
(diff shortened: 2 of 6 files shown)


assafgi:
* **#631** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/631?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/631?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

sergeyberezansky (approved):
I would rather leave the endpoints metrics (not remove it) and then open an other PR on making sure that we update the endpoints.

PR #636, opened 2025-12-18
fix: protect from non empty seed snapshot usage (CSI-400)

assafgi:


assafgi:
* **#636** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/636?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/636?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pkg/wekafs/volume.go
```diff
@@ -1149,6 +1149,56 @@ func (v *Volume) isFilesystemEmpty(ctx context.Context) (bool, error) {
...
+
+	fileNames, err := dir.Readdirnames(2)
+	if err == io.EOF {
+		logger.Debug().Msg("Seed snapshot is empty")
+		return true, nil
+	}
+
+	// Check if there are any files/dirs other than .snapshots
+	for _, name := range fileNames {
+		if name == SnapshotsSubDirectory {
```
sergeyberezansky: this is unnecessary, no .snapshots folder inside .snapshots/snapshot

sergeyberezansky:
This has been ported onto `dev` as #791, targeted at 2.9.2.

It is a behavioural port rather than a merge, since this branch is based on `main`: the port uses dev's `deferUmount` idiom, shares the directory scan with `isFilesystemEmpty` instead of duplicating it, and fails closed when the directory cannot be read (the original treated a read error as "empty", which would have waved through the case the check exists to catch).

Keeping this open until #791 merges.

sergeyberezansky:
Note: the port (#791) has been retargeted at **2.10** rather than 2.9.2, since refusing a seed snapshot that already contains data is a behaviour change for anyone relying on the current behaviour, not a pure fix.

sergeyberezansky:
Superseded by #791, which ports this fix onto the 2.10 line.

#791 carries the same guard - refusing to provision from a seed snapshot that has been written into - plus a test file this PR did not have, `deferUmount` in place of the bare deferred unmount, and the tracing boilerplate dropped. The directory check is shared with `isFilesystemEmpty` there as `dirHoldsOnlySnapshots`.

Two review comments on this PR were not carried across, deliberately: the `.snapshots` skip and the two-entry read are both needed for the filesystem-root caller, and sharing one helper keeps a branch that the seed-snapshot caller can never take. Harmless, and not worth splitting the helper over.

Closing in favour of #791. Thanks @assafgi - the fix is landing, just on the other line.

PR #640, opened 2026-01-07
fix: expose explicit backend connectivity metrics

assafgi:


assafgi:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/640?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#640** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/640?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/640?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#608** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/608?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sergeyberezansky:
Closing this in favour of the two pieces of it that are genuinely new, both of which have been ported onto the current stack. Thanks — the endpoint-rotation idea in particular is a real scale fix.

## Taken

**Endpoint rotation per request.** `rotateApiEndpointOnEachRequest`, propagated from a flag through the driver config to the API client. On by default for the metrics server, which polls continuously; off for the plugin, whose API traffic is sparse and where staying on one endpoint until it misbehaves keeps its per-endpoint failure counters meaningful.

One change from this PR's version: the rotation happens once at the top of `do()` rather than inside endpoint resolution. A single request resolves its endpoint twice — once to build the URL and again to record its metrics — so rotating on resolution attributes a request to a different endpoint than the one it was sent to, corrupting the metrics it feeds.

**Secret-name attribution.** `secret_name` is now a label on the per-volume metrics, as `namespace/name`, covering both the metrics server's capacity series and the controller's volume-health series.

Implemented from a different angle. This PR threaded a `secretName` parameter through `fromSecrets`/`GetClientFromSecrets` at thirteen call sites and passed `""` at every plugin one, because CSI hands the driver a Secret's contents and never its name. Instead the label is derived from the PersistentVolume, which does record the reference — and the shared label builder already receives the PV, so it covers both metric families with no call-site changes and no empty-string plumbing. It resolves through `preferredSecretRef`, the same ref both consumers already authenticate with, so the label always names the Secret actually in use.

## Not taken

**`weka_csi_metricsserver_backend_call`** duplicates `weka_csi_api_request_count`, which already counts every API call with `endpoint`, `method`, `url` and `status` labels. Shipping both would count each call twice under two names.

**The scrape counter.** It increments in `PersistentVolumeStreamer` around the PersistentVolume list fetch, on the same lines as the existing `fetch_pv_batch_operations_invoke_count` / `_success_count` / `_failure_count` / `_duration_seconds` family — so it reports what is already reported. Its `status_code` label is also hardcoded `"200"`/`"500"` rather than a real Kubernetes API status, which is misleading on a label of that name.

## One thing this surfaced

The API client maintains fifteen per-endpoint counters internally — failures, timeouts, per-status HTTP errors, transport/parse/no-response errors, request count and cumulative duration — which drive endpoint selection and rotation and are never exported to Prometheus. That is the gap the `backend_call` counter was reaching for, and exposing those directly would be the better version of this idea. Worth its own change.

PR #641, opened 2026-01-08
docs: add documentation about exposed metrics

assafgi:


docs/metrics-reference.md
```diff
@@ -0,0 +1,372 @@
...
+Track individual PV processing operations.
+
+| Metric Name | Type | Description |
+|-------------|------|-------------|
+| `weka_csi_metricsserver_process_pv_operations_count_total` | Counter | **Total PVs processed**. Count of individual PV processing operations completed. |
+| `weka_csi_metricsserver_process_pv_operations_duration_seconds` | Counter | **Cumulative PV processing duration**. Total time spent processing PVs. |
+| `weka_csi_metricsserver_process_pv_operations_duration_seconds_histogram` | Histogram | **PV processing duration distribution**. How long each PV takes to process. |
+
+### Metrics Fetching from Weka (Batch)
+
```
sergeyberezansky: Although we support it, but by default this is not used since we fetch quotas one by one

sergeyberezansky: The mode of metrics fetch is controlled by `metricsServer.enableBatchModeForQuotaUpdates`

docs/metrics-reference.md
```diff
@@ -0,0 +1,372 @@
...
+### Metrics Fetching from Weka (Single PV)
+
+Track individual PV metrics fetches.
+
+| Metric Name | Type | Description |
+|-------------|------|-------------|
+| `weka_csi_metricsserver_fetch_single_pv_metrics_invoke_count_total` | Counter | **Total single PV metrics fetches**. Count of individual PV metrics requests to Weka API. |
+| `weka_csi_metricsserver_fetch_single_pv_metrics_success_count_total` | Counter | **Successful single PV fetches**. Individual fetches that succeeded. |
+| `weka_csi_metricsserver_fetch_single_pv_metrics_failure_count_total` | Counter | **Failed single PV fetches**. Individual fetches that failed. Non-zero indicates volume or API issues. |
```
sergeyberezansky: or excessive load on WEKA cluster / API

docs/metrics-reference.md
```diff
@@ -0,0 +1,372 @@
...
+Track the periodic metrics collection scheduler.
+
+| Metric Name | Type | Description |
+|-------------|------|-------------|
+| `weka_csi_metricsserver_periodic_fetch_metrics_invoke_count_total` | Counter | **Total periodic fetch invocations**. How many times the periodic fetch timer has fired. |
+| `weka_csi_metricsserver_periodic_fetch_metrics_skip_count_total` | Counter | **Skipped periodic fetches**. Fetches skipped (e.g., previous fetch still running). High count may indicate fetch is slower than interval. |
+| `weka_csi_metricsserver_periodic_fetch_metrics_success_count_total` | Counter | **Successful periodic fetches**. Periodic fetches that completed successfully. |
+| `weka_csi_metricsserver_periodic_fetch_metrics_failure_count_total` | Counter | **Failed periodic fetches**. Periodic fetches that failed. Non-zero needs investigation. |
+
+### Quota Map Operations (Per Filesystem)
```
sergeyberezansky: Those are invoked only if batch quota fetches are performed

sergeyberezansky (changes requested):

sergeyberezansky:
Closing this — the documentation has been reworked and now lives on the current stack as `docs/prometheus-metrics.md`. Thanks; 91 of the 93 metrics documented here were exact matches, so this was the bulk of the work.

Note the diff shown above is misleading: `dev` was reset to `main` as part of moving development back onto `dev`, so this branch's base moved underneath it and the diff now shows all of the old divergence. The actual content was one file.

## What changed in the rework

**Organised by family rather than as one flat list.** There are five, and the original grouping put the read-only controller RPCs and volume health under the metrics server when both are emitted by the CSI plugin. Now: API client, CSI plugin service health, CSI plugin object metrics (volume health), metrics server service health, metrics server object metrics (per-volume capacity). Each section names the component that exposes it and the port to scrape.

**Ports** are taken from the Helm values with the key named, rather than from the Go defaults, and are no longer stated as one value shared by both components: `metrics.controllerPort` 9090, `metrics.nodePort` 9094, `metrics.metricsServerPort` 9096. `metricsServer.healthPort` 9196 serves the probes only and is not a metrics port.

**Corrections found against the code:**

- `weka_csi_api_endpoints_count` and `weka_csi_controller_delete_volume_total_capacity_bytes` do not exist. Removed.
- The four `weka_csi_volume_health_*` metrics, the read-only controller RPCs (`get_volume`, `list_volumes`, `validate_volume_capabilities`), `node_get_info` and `quota_map_miss_count_total` were missing. Added.
- The per-volume label table was missing `organization`, and now also carries `secret_name`.
- The six per-volume performance counters — reads, writes, and their byte and duration counterparts — are removed. They are declared and registered, but `PvStats.Performance` is never assigned anywhere, so the branch that would emit them is unreachable and they always report nothing. They should not be documented until the performance path is wired up.
- `free_bytes` was described as `capacity - used`. That is the arithmetic but not the meaning: it is headroom against the volume's *quota*, derived by the driver because a quota carries a hard limit rather than a used/free split. It says nothing about free space on the underlying filesystem, which can be exhausted while a volume still shows room.
- The total is 98, not 90.

**Added:** a dashboards section mapping each shipped dashboard and the alert rules to the metric families they read, and the freshness caveat stated where it belongs — per-volume capacity metrics carry the measurement timestamp rather than the scrape time, so any dashboard or alert window over them must exceed `quotaCacheValiditySeconds`.

`docs/monitoring.md` stays the operational guide and the two cross-reference; the PodMonitors are now documented there properly, including the `additionalLabels` trap where the object is created, Helm reports success, and Prometheus never scrapes it.

## Also worth knowing

The API client family is only two metrics, which reads as suspiciously small. It is correct — they are a CounterVec and a HistogramVec over six labels, so two names and many series — but the client additionally keeps fifteen per-endpoint counters that are never exported. That is noted in the new document, and is a candidate for a change of its own.

PR #643, opened 2026-01-14
fix(CSI-404): health check fails if any frontend is disconnected

caspx:
### TL;DR
  Health check now verifies ALL frontends are connected, not just the first one.

  ### What changed?
  - `isWekaRunning()` now calls `GetFrontendsFromDriver()` which parses all FE entries from `/proc/wekafs/interface`
  - Returns `false` if any frontend is disconnected
  - Added detailed logging showing which FEs are down
  - Added unit tests for the parsing logic

  ### How to test?
  1. Run unit tests: `go test ./pkg/wekafs/apiclient -run TestGetFrontendsFromDriver -v`
  2. Deploy CSI with a  client
  3. Stop one FE container
  4. Observe health check fails and pod restarts
  5. Check logs for `"Not all frontends are connected"` with `disconnected_frontends` field

  ### Why make this change?
  Previously, health check returned `true` if ANY frontend was connected, ignoring disconnected ones.

### DIFF
--- a/pkg/wekafs/apiclient/cluster.go
+++ b/pkg/wekafs/apiclient/cluster.go
@@ -5,15 +5,16 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/google/uuid"
-	"github.com/rs/zerolog/log"
-	"golang.org/x/exp/slices"
-	"k8s.io/apimachinery/pkg/util/rand"
 	"net"
 	"os"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/google/uuid"
+	"github.com/rs/zerolog/log"
+	"golang.org/x/exp/slices"
+	"k8s.io/apimachinery/pkg/util/rand"
 )
 
 const ApiPathLogin = "login"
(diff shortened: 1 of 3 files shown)


assafgi (approved):

pkg/wekafs/apiclient/cluster.go
```diff
@@ -325,18 +324,52 @@ func (a *ApiClient) getLocalContainersFromDriver(ctx context.Context) ([]ProcFsC
...
+			frontends = append(frontends, ProcFsContainer{
+				ContainerName: name,
+				ContainerId:   idInt,
+				Connected:     false,
+			})
+			continue
+		}
+
+		if !strings.Contains(line, "Connected frontend pid") {
+			continue
```
sergeyberezansky: This will cause to omit all containers that are not connected without appending to list.
So I guess that in such case we could have too short list of containers.
And part of containers that are not connected will not show up.
So later on, when you test disconnected > 0 you will miss those.

sergeyberezansky: IMHO, any thing that can be somehow parsed, should add up to the returned list of frontends

pkg/wekafs/utilities.go
```diff
@@ -563,25 +563,39 @@ func isWekaDriverLoaded() bool {
...
+	var disconnected []string
+	for _, frontend := range frontends {
+		if !frontend.Connected {
+			disconnected = append(disconnected, fmt.Sprintf("%s/FE%d", frontend.ContainerName, frontend.ContainerId))
 		}
-		log.Error().Msg("Client software is not running on host and NFS is not enabled")
 	}
-	return false
+
+	if len(disconnected) > 0 {
```
sergeyberezansky: due to the described above, here you will remain with disconnected=0 (because you just continue in many cases that should be counted as disconnected)

PR #656, opened 2026-04-22
feat: allow override of mount options using pod annotations

Gab-Menezes:
# Disclaimer
This is a POC and has not been tested because I don't have a spare cluster to test with.
This PR probably contains bugs and doesn't handle all edge cases correctly.

## Summary

This adds support for per-pod mount option overrides on via a pod annotation. When the CSI driver is deployed with `podInfoOnMount: true` on the `CSIDriver` object (which is already the case), Kubernetes injects pod identity information into every `NodePublishVolume` request. This takes advantage of that to allow workload owners to customize mount behavior at the pod level without touching the StorageClass or PVC.

Right now the mount options are stablished in the `StorageClass` at creation time, but maybe different volumes created using the same `StorageClass` might benefit from different mount options.

## How it works

When the feature is enabled (`--custompodmount`), the node server:
1. Reads the pod name and namespace from the CSI volume context (injected by Kubernetes)
2. Extracts the PV name from the kubelet target path and resolves the PVC claim name via the Kubernetes API
3. Fetches the pod and reads the `weka.io/mount-options` annotation
4. Matches the PVC claim name against the patterns in the annotation and applies the first match

## Annotation format

```yaml
metadata:
  annotations:
    weka.io/mount-options: |
      my-volume-.*: -forcedirect, +readcache
      my-vol-1: -forcedirect, +readcache, +writecache
      my-vol-2: -forcedirect
      my-vol-3: +inode_bits=64
```

Each line (or `;`-separated entry) maps a regex pattern against the PVC claim name to a set of mount option modifiers. `+option` adds an option, `-option` removes it. Mutually exclusive options (e.g. `writecache`/`readcache`/`coherent`) are handled correctly. PVCs not matched by any pattern are unaffected.

## Safety

- The feature is **opt-in** via `--custompodmount`; when disabled, there is zero overhead on the mount path and no Kubernetes API calls are made
- Mount options are applied after StorageClass and VolumeCapability flags, but **before** the readonly enforcement — `readOnly: true` on a PVC always wins and cannot be overridden by the annotation
- The node server ClusterRole is extended with `get` on `pods` to support the annotation lookup


pkg/wekafs/nodeserver.go
```diff
@@ -60,6 +63,7 @@ type NodeServer struct {
 	api               *ApiStore
 	config            *DriverConfig
 	semaphores        map[string]*semaphore.Weighted
+	kubeClient        kubernetes.Interface
```
sergeyberezansky, line 66: use a cached controller-runtime client ns.getConfig().GetDriver().manager.GetClient() instead.

pkg/wekafs/nodeserver.go
```diff
@@ -206,6 +239,12 @@ func NewNodeServer(nodeId string, maxVolumesPerNode int64, api *ApiStore, mounte
 		api:               api,
 		config:            config,
 		semaphores:        make(map[string]*semaphore.Weighted),
+		kubeClient: func() kubernetes.Interface {
```
sergeyberezansky, line 242: should not be necessary

pkg/wekafs/nodeserver.go
```diff
@@ -190,6 +194,35 @@ func getVolumeStats(volumePath string) (volumeStats *VolumeStats, err error) {
 	return &VolumeStats{capacityBytes, usedBytes, availableBytes, inodes, inodesUsed, inodesFree}, nil
 }
 
+// newKubeClient builds a Kubernetes client from in-cluster config or KUBECONFIG.
```
sergeyberezansky, line 197: should not be necessary

pkg/wekafs/nodeserver.go
```diff
@@ -327,6 +366,31 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+	mountFlags := req.GetVolumeCapability().GetMount().GetMountFlags()
+	volume.mountOptions.Merge(NewMountOptionsFromString(strings.Join(mountFlags, ",")), ns.getConfig().mutuallyExclusiveOptions)
+
+	// Apply per-pod mount option overrides from the weka.io/mount-options annotation.
+	// Requires podInfoOnMount: true on the CSIDriver so Kubernetes injects pod info into VolumeContext.
+	if ns.config.allowCustomPodMount && ns.kubeClient != nil && params != nil {
+		podName := params[CSIPodNameKey]
+		podNamespace := params[CSIPodNamespaceKey]
+		if podName != "" && podNamespace != "" {
+			pvcName, pvErr := getPVCClaimNameFromTargetPath(ctx, ns.kubeClient, targetPath)
```
sergeyberezansky, line 380: PVC claim name is not necessarily correct in the path, since path uses sanitized names, isn't it?
Although usually it is in form of pvc-<uid>, the conformance to CSI standard should allow other names as well. 
I am double-checking this

pkg/wekafs/podannotation.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	return pv.Spec.ClaimRef.Name, nil
+}
+
+// getPodAnnotationMountOptions fetches the pod, reads the weka.io/mount-options annotation,
+// and returns the raw mount option modifiers for the first pattern that matches pvcClaimName.
+// Returns "" if the annotation is absent or no pattern matches.
+func getPodAnnotationMountOptions(ctx context.Context, kubeClient kubernetes.Interface, podNamespace, podName, pvcClaimName string) string {
+	logger := log.Ctx(ctx)
+	pod, err := kubeClient.CoreV1().Pods(podNamespace).Get(ctx, podName, metav1.GetOptions{})
```
sergeyberezansky, line 127: should use crclient

pkg/wekafs/podannotation.go
```diff
@@ -0,0 +1,150 @@
...
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	// PodMountOptionsAnnotation is the annotation key on pods for per-PVC mount option overrides.
+	// Format: one entry per line (or separated by ';'):
+	//   <pvc-name-regex>: <mount-option-modifiers>
+	// Mount option modifiers are comma-separated with + (add) or - (remove) prefix.
+	// Example:
+	//   my-volume-.*: -forcedirect, +readcache
```
sergeyberezansky, line 19: in most cases, mountOptions could be added without modifieds, e.g. writecache (this will both override other option value if exist and remove the mutually exclusive ones)
Only -option makes sense
It is good to have both options but documentation or comments should also incorporate this

pkg/wekafs/podannotation.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/rs/zerolog/log"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	// PodMountOptionsAnnotation is the annotation key on pods for per-PVC mount option overrides.
+	// Format: one entry per line (or separated by ';'):
+	//   <pvc-name-regex>: <mount-option-modifiers>
+	// Mount option modifiers are comma-separated with + (add) or - (remove) prefix.
```
sergeyberezansky, line 17: with optional + (force add) or - (force remove)

pkg/wekafs/nodeserver.go
sergeyberezansky: Why do you need to get the PVC claim name from target path?
```
vc := req.GetVolumeContext()
pvName := vc["csi.storage.k8s.io/pv/name"]
pvcName := vc["csi.storage.k8s.io/pvc/name"]
```

pkg/wekafs/podannotation.go
```diff
@@ -0,0 +1,150 @@
...
+			toAdd := NewMountOptionsFromString(part)
+			opts.Merge(toAdd, exclusives)
+		}
+	}
+	return opts
+}
+
+// extractPVNameFromTargetPath parses the PV name out of a kubelet CSI target path.
+// Expected format: .../volumes/kubernetes.io~csi/<pvName>/mount
+func extractPVNameFromTargetPath(targetPath string) string {
```
sergeyberezansky, line 95: see the comment above. The necessary info should be already in VolumeContext

pkg/wekafs/podannotation.go
```diff
@@ -0,0 +1,150 @@
...
+		if part == "kubernetes.io~csi" && i+1 < len(parts) {
+			return parts[i+1]
+		}
+	}
+	return ""
+}
+
+// getPVCClaimNameFromTargetPath extracts the PV name from the target path and
+// returns the PVC claim name bound to that PV.
+func getPVCClaimNameFromTargetPath(ctx context.Context, kubeClient kubernetes.Interface, targetPath string) (string, error) {
```
sergeyberezansky, line 107: see the comment above. The necessary info should be already in VolumeContext

sergeyberezansky (changes requested):

sergeyberezansky:
I would further expand it by allowing PVC annotations as well, with very strict order:
StorageClass > PVC > Pod

sergeyberezansky:
The feature is implemented in https://github.com/weka/csi-wekafs/pull/678
Will be merged when testing is finished.

PR #682, opened 2026-05-26
fix: use unique port names

assafgi:


sergeyberezansky:
Reworked onto `dev` as #792, so the fix ships with 2.9.2.

Two changes from this branch:

- **Shorter names.** #792 uses `ns-healthz` / `reg-healthz` / `at-metrics` rather than `register-healthz` / `attacher-metrics` / `provisioner-metrics`. Those short forms are already what the 3.0 branch uses, where the same collision was fixed independently as part of CSI-387 and the metrics work — so rendering `dev` and the 3.0 tip now produces identical port names, and no second set of names propagates behind the fix.
- **Only the attacher is renamed** on the controller. The provisioner keeps `pr-metrics`, since renaming just one side is enough to break the collision.

Verified by rendering the chart both ways: `dev` today reports duplicate `healthz` and `pr-metrics`, #792 reports none.

Closing this one in favour of #792 — thanks, the diagnosis was correct and is shipping.

PR #692, opened 2026-06-09
fix: remove nil-apiClient legacy mode, error on missing API credentials

kristina-solovyova:
### TL;DR

Removed legacy "API-less" volume operation mode, making an API client mandatory for all volume operations.

### What changed?

- `GetClientFromSecrets` now returns an error instead of `nil, nil` when no API credentials are available, with actionable error messages guiding users to configure a StorageClass secret or global API secret.
- Mount operations now always require an API client to obtain a mount token; the legacy codepath that skipped token retrieval when no client was present has been removed.
- `pruneUnsupportedMountOptions` no longer drops `sync_on_close` when the API client is absent — the option is only dropped when the cluster explicitly reports it as unsupported.
- `getMaxCapacity` and `ensureSufficientFsSizeOnUpdateCapacity` no longer have special-case branches for a missing API client.
- `UpdateCapacity` no longer falls back to xattr-based capacity tracking when no API client is bound.
- `isEncrypted` now safely returns `false, nil` when the encrypted field is unset, preventing a nil pointer dereference.
- `NewVolumeForBlankVolumeRequest` now wraps the API client retrieval error as a proper gRPC `FailedPrecondition` status error.
- The `CanBeOperated` doc comment was corrected to accurately describe its return semantics.

### Why make this change?

The legacy mode allowed volumes to be operated without an API client, relying on xattr-based capacity tracking and unauthenticated mounts. This mode is no longer supported and its presence was masking misconfigurations by silently falling back to degraded behavior. Making the API client mandatory ensures failures are surfaced early with clear, actionable error messages rather than silently producing volumes that behave incorrectly.

### DIFF
--- a/pkg/wekafs/volume.go
+++ b/pkg/wekafs/volume.go
@@ -108,9 +108,6 @@ func (v *Volume) pruneUnsupportedMountOptions(ctx context.Context) {
 		if v.apiClient != nil && !v.apiClient.SupportsSyncOnCloseMountOption() {
 			logger.Debug().Str("mount_option", MountOptionSyncOnClose).Msg("Mount option not supported by current Weka cluster version and is dropped.")
 			v.mountOptions = v.mountOptions.RemoveOption(MountOptionSyncOnClose)
-		} else if v.apiClient == nil {
-			logger.Debug().Str("mount_option", MountOptionSyncOnClose).Msg("Cannot determine current Weka cluster version, dropping mount option.")
-			v.mountOptions = v.mountOptions.RemoveOption(MountOptionSyncOnClose)
 		}
 	}
 	if v.mountOptions.hasOption(MountOptionReadOnly) {
@@ -218,6 +215,9 @@ func (v *Volume) isEncrypted(ctx context.Context) (bool, error) {
 				}
 			}
 		}
+	}
+	if v.encrypted == nil {
+		return false, nil
 	}
 	return *v.encrypted, nil
 }
@@ -474,12 +474,6 @@ func (v *Volume) getFilesystemTotalCapacity(ctx context.Context) (int64, error) 
 }
 
 func (v *Volume) getMaxCapacity(ctx context.Context) (int64, error) {
-
-	if v.apiClient == nil {
-		// this is a legacy, API-unbound volume
-		return v.getFilesystemFreeSpace(ctx)
-	}
-
 	// max size of the volume is the current size of the filesystem (or 0 if not exists) + free space on storage
 	currentFsSize, err := v.getFilesystemTotalCapacity(ctx)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


kristina-solovyova:
* **#692** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/692?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/692?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

assafgi (approved):

kristina-solovyova:
@sergeyberezansky is it ok to merge this one?

pkg/wekafs/volume.go
```diff
@@ -523,7 +517,6 @@ func (v *Volume) getCapacityFromQuota(ctx context.Context) (capacity int64, retE
 			return int64(size), nil
 		}
 	}
-	logger.Trace().Msg("Volume appears to be a legacy volume, failing back to Xattr")
 	size, err := v.getSizeFromXattr(ctx)
```
sergeyberezansky, line 520: This flow becomes outdated and should be removed

pkg/wekafs/wekafs.go
```diff
@@ -278,8 +278,9 @@ func (api *ApiStore) GetClientFromSecrets(ctx context.Context, secrets map[strin
 			logger.Trace().Msg("No explicit API service for request, using legacySecrets")
 			secrets = *api.legacySecrets
 		} else {
-			logger.Trace().Msg("No API service for request, switching to legacy mode")
-			return nil, nil
+			logger.Warn().Msg("No API credentials provided for this request and no global API secret is configured")
+			return nil, fmt.Errorf("no API credentials provided for this request and no global API secret is configured; " +
+				"add csi.storage.k8s.io/provisioner-secret-name to the StorageClass or configure a global API secret")
```
sergeyberezansky, line 283: We should remove global secret as well

sergeyberezansky (changes requested):
@kristina-solovyova if you take a look at the 3.0 branch, you will see that legacy removal is bigger.
It also includes removal of xattr flow, as well as removal of dev mode.
and modification of sanity to use nfs only.

sergeyberezansky:
Closing as superseded by #785, which removes the API-less legacy paths as part of CSI-314 and goes further: it also drops `debugPath`, dev mode, the directory-structure mimic, and the `legacyVolumesApiSecret`, with the breaking change documented for operators.

The CI change here is moot on `dev` — `.github/workflows/sanity.yaml` no longer exists, so there is no `legacy_sanity` job left to remove.

One piece of this PR is worth keeping and is being taken separately: the `isEncrypted` nil guard. That is a real bug and it is present on both `dev` and the CSI-314 branch — `v.encrypted` is still nil when `Exists()` returns an error, so `return *v.encrypted` panics, and `isEncryptedSafe` swallows the error but not the panic. Your two-line guard is the fix; it will land with credit.

Thanks — the diagnosis here was right, it is just arriving through the CSI-314 work.

PR #693, opened 2026-06-17
fix: preserve thin FS on FS expansion

assafgi:


assafgi:
* **#693** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/693?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/693?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pkg/wekafs/volume.go
```diff
@@ -578,6 +578,18 @@ func (v *Volume) resizeFilesystem(ctx context.Context, capacity int64) error {
 	capLimit := capacity
 
 	fsu := apiclient.NewFileSystemResizeRequest(fsObj.Uid, &capLimit)
+	if fsObj.IsThinlyProvisioned {
+		// Weka reverts a filesystem to thick provisioning if SSD params are omitted on resize.
+		// Preserve the existing thin configuration so an auto-expand does not silently convert it.
+		minSsd := fsObj.ThinProvisioningMinSsd
+		fsu.ThinProvisionMinSsd = &minSsd
+		// thin max must always equal the total capacity
+		maxSsd := capLimit
```
sergeyberezansky: I think that this is incorrect assumption.  
Maybe that could be configurable and we could provide few options:

1\. Do not change (resize filesystem but do not increase ssd size)  
2\. IncreaseBySize (resize ssd by adding the exact capacity which basically means again convert it to thick) and this is not good.  
3\. IncreaseByPercentage (suppose previous FS is 1TB, and SSD max is 100GB, then resize to 2TB should render the new max_ssd to be 100* (2/1)

and 3 should be default  
because current solution presumably leaves the fs thin but sets the SSD max to actual FS size, effectively making it thick again

sergeyberezansky (changes requested):
the suggested solution is incorrect.

suppose you have 1TB filesystem and max_ssd_capacity set to 100GB.  
then you resize filesystem to 2TB, and it will set max_ssd_capacity to 2TB (in fact make it thick again)  
Instead we should either not touch the ssd allocation at all, or to scale it relatively.

sergeyberezansky (approved):

PR #707, opened 2026-07-31
fix: bump google.golang.org/grpc to v1.82.1 (dependabot alert #24)

assafgi:


sergeyberezansky:
Closing as superseded — both `main` and `dev` are already on `google.golang.org/grpc v1.83.0`, so merging this would be a downgrade to v1.82.1.

Dependabot alert #24 is addressed: v1.82.1 was the fix version and v1.83.0 is ahead of it.

Thanks for raising it — the bump landed through another route in the meantime.

PR #714, opened 2026-08-08
fix: invalidate filesystem cache before deletion to prevent stale UID snapshot listing

sergeyberezansky:
### TL;DR
Deleting a filesystem could use stale cached data, letting deletion proceed even if the filesystem still had snapshots.

### What changed?
- The snapshot check performed before deleting a filesystem now always reads current data instead of a cached copy.
- The cached record for a filesystem is cleared immediately once its deletion begins.

### How to test?
1. This depends on an internal timing window that isn't practical to reproduce manually.
2. Covered by automated tests added in this change.

### Why make this change?
A cache used to avoid extra lookups wasn't cleared at the right time, so a delete could occasionally act on outdated filesystem information.

PR #713, opened 2026-08-08
fix: default semaphore weight to 1 for ops absent from maxConcurrencyPerOp

sergeyberezansky:
### TL;DR
Operations without an explicit concurrency limit were being blocked entirely instead of allowed to run.

### What changed?
- Operations missing a configured concurrency limit now default to allowing 1 at a time, instead of 0.
- Operations explicitly configured with a limit of 0 still remain blocked, as intended.

### How to test?
1. Deploy the CSI driver.
2. Trigger an operation that has no explicit concurrency limit configured.
3. Verify it completes normally instead of hanging until it times out.

### Why make this change?
An operation without a configured limit defaulted to a concurrency of zero, which can never be acquired, so it would hang until its request timed out rather than running or failing clearly.

PR #712, opened 2026-08-08
fix: replace Mutex with RWMutex in ApiStore to prevent concurrent map access races

sergeyberezansky:
### TL;DR
Fixes a bug that could crash the CSI controller under concurrent use

### What changed?
- Fixed unsafe concurrent access to the driver's internal cache of Weka cluster connections
- Prevents a crash when many requests need cluster credentials at the same time

### How to test?
1. Covered by automated concurrency tests included in this change
2. Before the fix, the symptom was occasional controller pod restarts under heavy simultaneous volume operations; this should no longer occur

### Why make this change?
The controller could crash outright when multiple requests needing storage-cluster credentials were handled at the same time — more likely at scale or with many simultaneous volume operations.

PR #711, opened 2026-08-08
refactor: split `wekafs.go` into focused files and fix import grouping

sergeyberezansky:
### TL;DR
Pure internal code reorganization — no behavior change

### What changed?
- Split internal driver code into smaller, more focused files
- Cleaned up code formatting

### How to test?
1. Deploy and use the driver as before
2. Confirm behavior is identical to prior releases (verified by automated tests)

### Why make this change?
Makes the codebase easier to maintain and extend, in preparation for upcoming feature work.

PR #715, opened 2026-08-08
feat: add paginated API response support for quota fetching

sergeyberezansky:
### TL;DR
Quota lists are no longer silently truncated when a filesystem has more quotas than fit in one page of results.

### What changed?
- The API client now follows the storage system's pagination and combines all pages into the full result.
- Hardened handling for older clusters without pagination support, and for empty or malformed pages.

### How to test?
1. Query quotas on a filesystem with more quota entries than fit in a single page.
2. Verify the full list is returned, not just the first page.
3. Also covered extensively by automated tests.

### Why make this change?
The client wasn't following the "next page" pointer the storage system returns for long lists, so any sufficiently long list - most noticeably quotas - was silently incomplete.

### DIFF
--- a/pkg/wekafs/apiclient/requests.go
+++ b/pkg/wekafs/apiclient/requests.go
@@ -239,12 +239,24 @@ func (a *ApiClient) request(ctx context.Context, Method string, Path string, Pay
 				apiError: reqErr,
 			}
 		}
-		err := json.Unmarshal(rawResponse.Data, v)
-		if err != nil {
-			logger.Error().Err(err).Interface("object_type", reflect.TypeOf(v)).Msg("Failed to marshal JSON request into a valid interface")
+		unmarshalErr := json.Unmarshal(rawResponse.Data, v)
+		if unmarshalErr != nil {
+			logger.Error().Err(unmarshalErr).Interface("object_type", reflect.TypeOf(v)).Msg("Failed to marshal JSON request into a valid interface")
 		}
 		switch s {
 		case http.StatusOK:
+			if unmarshalErr != nil {
+				// A page whose data fails to decode must not be folded in as an empty page -
+				// that would silently truncate a paginated result. Only the success path can
+				// reach here relying on unmarshalErr; non-200 paths return their own reqErr
+				// below and never look at it.
+				return ApiNonTransientError{apiError: ApiError{
+					Err:         unmarshalErr,
+					Text:        "Failed to unmarshal JSON response body",
+					StatusCode:  s,
+					ApiResponse: rawResponse,
+				}}
+			}
 			nextToken = rawResponse.NextToken
 			return nil
 		case http.StatusUnauthorized:
@@ -280,7 +292,11 @@ func (a *ApiClient) Request(ctx context.Context, Method string, Path string, Pay
 		return err
 	}
 
+	// Only a GET can safely be repeated per page: a POST/PUT/DELETE whose response type happens
+	// to implement ApiObjectResponse would otherwise re-send the same mutating payload once per
+	// page.
 	accumulated, paginated := Response.(ApiObjectResponse)
+	paginated = paginated && Method == http.MethodGet
 	if !paginated {
 		_, err := a.request(ctx, Method, Path, Payload, Query, Response)
 		if err != nil {
@@ -293,12 +309,45 @@ func (a *ApiClient) Request(ctx context.Context, Method string, Path string, Pay
 	if responseType.Kind() != reflect.Ptr {
 		return errors.New("a paginated response must be passed as a pointer")
 	}
+	responseValue := reflect.ValueOf(Response)
+	if responseValue.IsNil() {
+		// accumulated wraps a nil *T here: the interface value itself is non-nil (it carries a
+		// type), so the type assertion above succeeds, but CombinePartialResponse would
+		// dereference a nil receiver and panic. Fail cleanly instead.
+		return fmt.Errorf("a paginated response must not be a nil %s pointer", responseType.Elem())
+	}
+	logger := log.Ctx(ctx)
+
+	if !a.SupportsUrlQueryParams() {
+		// The cluster can't accept next_token at all, so pagination cannot happen: every
+		// request would come back as page 1 again (see the non-advancing-token check below,
+		// which exists for exactly this shape of backend misbehaviour). Do a single request and
+		// return it, matching the pre-pagination behaviour, rather than looping or failing a
+		// call that used to succeed.
+		logger.Warn().Str("path", Path).Msg("Cluster does not support URL query parameters; fetching a single page, result may be truncated")
+		_, err := a.request(ctx, Method, Path, Payload, Query, Response)
+		return err
+	}
+
 	// Work on a copy: Query belongs to the caller, and it is legitimately nil, which Set would
 	// panic on.
 	query := cloneQuery(Query)
-	logger := log.Ctx(ctx)
+
+	// Zero the caller's accumulator before fetching the first page. CombinePartialResponse
+	// appends, so a caller that passes in a non-empty accumulator (or reuses one across calls)
+	// would otherwise get duplicates alongside the freshly fetched pages. This restores the
+	// replace-semantics of the pre-pagination code path, which unmarshalled straight into
+	// Response.
+	responseValue.Elem().Set(reflect.Zero(responseType.Elem()))
+
+	// currentToken is the next_token that produced the page just received - used to detect a
+	// backend that echoes the same token back forever instead of advancing it.
+	currentToken := ""
 
 	for page := 1; ; page++ {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
 		// Every page is unmarshalled into a fresh object and then folded into the caller's
 		// Response. Unmarshalling straight into Response would let each page overwrite the last.
 		pageValue := reflect.New(responseType.Elem()).Interface()
@@ -316,10 +365,14 @@ func (a *ApiClient) Request(ctx context.Context, Method string, Path string, Pay
 			}
 			return nil
 		}
+		if nextToken == currentToken {
+			return fmt.Errorf("backend returned pagination token %q unchanged after it was already used for %s; refusing to loop indefinitely since the backend is not advancing the token", nextToken, Path)
+		}
 		if page >= ApiMaxPagesPerRequest {
 			return fmt.Errorf("paginated response exceeded %d pages, giving up", ApiMaxPagesPerRequest)
 		}
 		query.Set("next_token", nextToken)
+		currentToken = nextToken
 	}
 }
 
--- a/pkg/wekafs/apiclient/requests.go
+++ b/pkg/wekafs/apiclient/requests.go
@@ -205,14 +205,18 @@ func (a *ApiClient) do(ctx context.Context, Method string, Path string, Payload 
 	}
 }
 
-// request wraps do with retries and some more error handling
-func (a *ApiClient) request(ctx context.Context, Method string, Path string, Payload *[]byte, Query url.Values, v interface{}) apiError {
+// request wraps do with retries and some more error handling. It returns the token of the next
+// page when the backend indicates the result was truncated, or an empty string when it was not.
+func (a *ApiClient) request(ctx context.Context, Method string, Path string, Payload *[]byte, Query url.Values, v interface{}) (string, apiError) {
 	op := "ApiClientRequest"
 	ctx, span := otel.Tracer(TracerName).Start(ctx, op)
 	defer span.End()
 	ctx = log.With().Str("span_id", span.SpanContext().SpanID().String()).Logger().WithContext(ctx)
 	logger := log.Ctx(ctx)
+	var nextToken string
 	f := func() apiError {
+		// Reset per attempt: a retry must not inherit the token of an attempt that later failed.
+		nextToken = ""
 		rawResponse, reqErr := a.do(ctx, Method, Path, Payload, Query)
 		if a.handleTransientErrors(ctx, reqErr) != nil { // transient network errors
 			a.rotateEndpoint(ctx)
@@ -241,6 +245,7 @@ func (a *ApiClient) request(ctx context.Context, Method string, Path string, Pay
 		}
 		switch s {
 		case http.StatusOK:
+			nextToken = rawResponse.NextToken
 			return nil
 		case http.StatusUnauthorized:
 			logger.Warn().Msg("Got Authorization failure on request, trying to re-login")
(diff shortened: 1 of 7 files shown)


sergeyberezansky:
* **#741** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/741?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#740** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/740?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#733** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/733?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#726** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/726?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#725** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/725?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#724** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/724?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#723** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/723?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#721** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/721?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#751** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/751?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#750** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/750?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#720** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/720?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#719** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/719?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#718** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/718?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#717** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/717?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#716** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/716?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#715** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/715?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/715?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#714** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/714?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#713** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/713?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#712** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/712?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#711** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/711?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#710** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/710?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#709** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/709?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#708** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/708?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

kristina-solovyova (approved):

kristina-solovyova (changes requested):
🤖 **Automated code review by Claude** (posted via `gh` by kristina-solovyova)

Reviewed commit `ccc72a56`. Verified locally: `gofmt -l` and `go vet` clean, the four new pagination tests pass (also under `-race`). `TestApiClientRequest` fails, but that is pre-existing — it requires a live cluster at `localhost:14000`, unrelated to this diff.

## Must fix

### 1. Pagination degrades into a 1000-request spin on clusters without URL query param support — `pkg/wekafs/apiclient/requests.go:301`

`do()` applies query params only when the cluster supports them:

```go
if Query != nil && len(Query) > 0 && a.SupportsUrlQueryParams() {
    r.URL.RawQuery = Query.Encode()
}
```

`CompatibilityMap` is zero-valued at construction (`apiclient.go:90`), and `UrlQueryParams` is explicitly set to `false` both for clusters below the minimum version and whenever the version string fails to parse (`compatibility.go:80,118`).

In those cases `query.Set("next_token", nextToken)` is silently dropped on the wire. The backend returns page 1 along with the same token, so the loop issues 1000 identical requests, appends 1000 duplicate copies of page 1 into the caller's `*Quotas`, and then fails with `paginated response exceeded 1000 pages`. Before this change the same call made exactly one request and succeeded (truncated).

Suggested fix: gate the paginated path on `a.SupportsUrlQueryParams()` and fall back to a single request, or fail fast with a clear error.

### 2. No detection of a non-advancing token — `pkg/wekafs/apiclient/requests.go:319`

Nothing compares `nextToken` against the token used for the previous page. Any backend that echoes an unchanged token — or ignores `next_token` entirely, see #1 — costs 1000 requests, each with up to `ApiRetryMaxCount` retries and exponential backoff, before the error surfaces.

Breaking out when the new token equals the previous one turns several minutes of hammering the cluster into an immediate, diagnosable stop. It is also a cheap second line of defense for #1.

### 3. Unmarshal failure is logged but not returned, and pagination now amplifies it — `pkg/wekafs/apiclient/requests.go:242`

```go
err := json.Unmarshal(rawResponse.Data, v)
if err != nil {
    logger.Error().Err(err).Interface("object_type", reflect.TypeOf(v)).Msg("Failed to marshal JSON request into a valid interface")
}
switch s {
case http.StatusOK:
    nextToken = rawResponse.NextToken
    return nil
```

A page whose `data` fails to decode (`data: null`, or a shape change) is folded in as an empty page, and the loop keeps following tokens. The caller receives a silently short result set with no error at all. Previously a single bad response at least produced an empty result the caller could notice; now missing pages in the middle of a set are invisible.

Please return the unmarshal error, at minimum for the paginated path.

## Worth fixing

### 4. The loop never checks `ctx`, and cancellation cannot reach the HTTP layer — `requests.go:305`

`do()` builds requests with `http.NewRequest`, not `NewRequestWithContext`, so a cancelled or expired `ctx` does not abort in-flight requests. Combined with up to 1000 iterations × (60s `ApiHttpTimeOutSeconds` + retries), a runaway pagination keeps a controller goroutine issuing API calls long after the gRPC caller (e.g. `CreateVolume`) has timed out and the sidecar has retried. A `ctx.Err()` check at the top of each iteration is enough here.

### 5. `CombinePartialResponse` appends into the caller's object without clearing it — `quota.go:96`

The old path (`json.Unmarshal` straight into `Response`) replaced the slice contents; the new path appends. A caller that reuses a non-empty `*Quotas` across two `Get` calls, or retries at a higher level with the same accumulator, now gets duplicated entries — which for `GetQuotaByFilter` would turn a single match into `MultipleObjectsFoundError`.

Related: when page N fails, `Request` returns an error but the caller's object is left holding pages 1..N-1. No current caller does this, but the contract is a footgun — consider resetting the accumulator on page 1, or documenting the append semantics explicitly.

### 6. A typed-nil pointer response now panics instead of erroring — `requests.go:283`

A nil `*Quotas` stored in the `Response interface{}` satisfies `ApiObjectResponse`, so the paginated branch is taken and `accumulated.CombinePartialResponse(...)` executes `*q = append(*q, …)` on a nil receiver — a nil-pointer dereference. The old code path handed the same value to `json.Unmarshal`, which returned an `InvalidUnmarshalError` that was merely logged. A nil check next to the existing pointer-kind check closes this.

## Effectiveness gap

### 7. Pagination is wired only into a function with no production callers — `quota.go:288`

`Quotas` is the sole type implementing `ApiObjectResponse`, and its only consumer is `FindQuotaByFilter`, called only by `GetQuotaByFilter` — and neither is referenced anywhere outside `pkg/wekafs/apiclient` (only `quota.go` and `pagination_test.go`).

Meanwhile the list calls actually used at runtime still fetch a single page and remain truncatable:

- `FindSnapshotsByFilter` — `snapshot.go:45`
- `FindNfsPermissionsByFilter` / `FindNfsPermissionsByFilesystem` — `nfs.go:143,162`
- filesystem lists — `filesystem.go:102,118`, NFS permissions on filesystems `filesystem.go:236`
- interface groups — `interfacegroup.go:123,148`

On a cluster with many snapshots, `GetSnapshotByFilter` still silently misses objects — the exact class of bug the preceding commit (`53e0e5b3`, snapshot listing) was about. Converting at least `[]Snapshot` and `[]NfsPermission` to paginated types would make the feature deliver its intent.

### 8. Minor: pagination is applied to every HTTP method, including mutating ones — `requests.go:277`

`Request` decides purely on the response type, so a `POST`/`PUT`/`DELETE` whose response type implements `ApiObjectResponse` would be re-sent with the identical `Payload` plus `next_token`, re-executing the mutation once per page. No such type exists today, but restricting the paginated path to `GET` removes the trap for the next type that opts in.

sergeyberezansky:
Addressed in `6223c2f4`. Six of the eight findings fixed; #7 dismissed with reasoning below.

## Fixed

**1 — pagination degrading into a 1000-request spin.** Confirmed, and slightly worse than described: `SupportsUrlQueryParams()` is false not only below v4.0 but also whenever the cluster's version string **fails to parse** (`compatibility.go:80`), so an unexpected version format was enough to trigger it. The paginated path now short-circuits to a single request with a warning that the result may be truncated — deliberately preserving the pre-pagination behaviour rather than failing a call that used to succeed.

**2 — non-advancing token.** Now compared against the token just sent; stops immediately with an error naming the path.

**3 — unmarshal error swallowed.** Returned now, on the success path only. The non-200 branches still rely on `reqErr` exactly as before, since the unmarshal runs ahead of the status switch.

**4 — `ctx` unchecked in the loop.** `ctx.Err()` checked each iteration. Left `do()` on `http.NewRequest` — switching to `NewRequestWithContext` is a broader change than this PR should carry.

**5 — accumulator appended instead of replaced.** Zeroed before page 1, restoring the replace semantics of the original `json.Unmarshal`. Done via reflection rather than adding a `Reset()` to `ApiObjectResponse`: a third implementer (`QuotaListResponse`) arrives in #722 above this PR, and changing the interface here would have broken it and caused restack conflicts.

**6 — typed-nil pointer panic.** Nil check beside the existing pointer-kind check.

**8 — pagination applied to mutating methods.** Restricted to `GET`.

## Dismissed

**7 — "pagination is wired only into a function with no production callers."** The premise is incorrect: `Quotas` is not the sole implementer — `FileSystems` implements `ApiObjectResponse` too (`filesystem.go:146`). More importantly the conclusion doesn't hold, because **#722 adds `QuotaListResponse`** (`quotamap.go:57`), and `GetQuotaMap` — the bulk fetch of every quota on a filesystem — passes it through this exact path. That is the one place where pages genuinely occur in production. Filesystem, snapshot and interface-group listings fit in a single page in practice, so converting them would add machinery that never engages.

This is worth flagging for future reviews of this stack: the finding was correct about the diff in isolation and wrong about the system, because the consumer lives in a PR *above* this one. Reviewing a stacked PR needs the upstack context, not just its own diff.

One unrelated thing surfaced while checking #7, not fixed here: `FileSystems.CombinePartialResponse` can never execute, because `filesystem.go:170` passes `(*[]FileSystem)(fsList)` and that conversion strips the method set, so the value reaching `Request` doesn't satisfy the interface. Harmless today given filesystems fit one page, but it's dead code that reads as wired up.

## Testing

Seven new tests, one per fix, each verified to fail when that fix alone is reverted. Findings 1 and 3 were checked by re-running with the fix disabled: with #1 off the trace shows the backend returning `next_token: "page-1"` unchanged on consecutive requests — the exact runaway. Note #1 and #2 overlap: with #1 disabled, #2's guard turns the 1000-request spin into a fast error, but only #1 keeps such clusters working at all.

`gofmt` and `go vet` clean; the apiclient suite passes. `TestApiClient_ResolvePathToInode` still fails without a live cluster at `localhost:14000` — pre-existing and unrelated.

pkg/wekafs/apiclient/requests.go
```diff
@@ -235,12 +239,25 @@ func (a *ApiClient) request(ctx context.Context, Method string, Path string, Pay
...
-		err := json.Unmarshal(rawResponse.Data, v)
-		if err != nil {
-			logger.Error().Err(err).Interface("object_type", reflect.TypeOf(v)).Msg("Failed to marshal JSON request into a valid interface")
+		unmarshalErr := json.Unmarshal(rawResponse.Data, v)
+		if unmarshalErr != nil {
+			logger.Error().Err(unmarshalErr).Interface("object_type", reflect.TypeOf(v)).Msg("Failed to marshal JSON request into a valid interface")
 		}
 		switch s {
 		case http.StatusOK:
+			if unmarshalErr != nil {
```
kristina-solovyova, line 255: This strictness applies to **every** call, not just paginated pages — `do`'s status switch is shared by the whole client.

All DELETEs pass `&ApiResponse{}` as the response target (`quota.go:399`, `filesystem.go:216`, `snapshot.go:151`, `nfs.go:324`, `nfs.go:564`). Those 200 bodies carry no `data` key, so `ApiResponse.Data` is nil and `json.Unmarshal` returns `unexpected end of JSON input` — previously logged and ignored, now a permanent `ApiNonTransientError`. That would turn a completed delete into a hard failure and make the sidecar retry forever.

Same bluntness inside the pagination loop: a legitimately empty page (absent `data`) is indistinguishable from malformed `data`, so an empty result set fails the whole call.

Suggest distinguishing the two rather than gating on the response type:

```go
case http.StatusOK:
    if len(rawResponse.Data) == 0 {
        // No data key at all: an empty page, or a body-less success such as DELETE.
        // Nothing to fold in, and nothing malformed either.
        nextToken = rawResponse.NextToken
        return nil
    }
    if unmarshalErr != nil {
        // Data was present but did not decode - folding this in as an empty page
        // would silently truncate a paginated result.
        return ApiNonTransientError{...}
    }
```

Keeps the anti-truncation property the comment is after, and unbreaks body-less successes.

Not verified against a live cluster — whether Weka's DELETEs and empty paginated pages actually omit `data` needs a real endpoint. The `len(...) == 0` guard is correct either way.

sergeyberezansky: Confirmed and fixed in `d0f6b164`. You're right, and this was a regression I introduced — worse than the truncation it was meant to prevent.

Verified the mechanism rather than reasoning about it: unmarshalling a `{"status":"OK"}` body gives `rawResponse.Data == nil, len 0`, and `json.Unmarshal(nil, target)` returns `unexpected end of JSON input`. So every one of the five DELETE sites you listed would have turned a delete that already succeeded on the cluster into a permanent `ApiNonTransientError`, with the sidecar retrying it forever.

Took your suggested shape — `len(rawResponse.Data) == 0` first, `unmarshalErr` second — so *nothing to decode* and *present but malformed* are distinguished, and the anti-truncation property survives. One addition: the unmarshal itself is now skipped when there is no data, so the error log no longer fires on every delete either.

Two tests, both mutation-verified (they fail with exactly `unexpected end of JSON input` when the guard is removed):

- a DELETE-shaped body-less 200 returns no error
- a paginated fetch whose final page carries no `data` completes with the earlier pages intact, in 2 requests

On your caveat about not having verified against a live cluster: agreed, and worth stating plainly that neither of us has. The guard is correct whichever way Weka actually frames those bodies, but if the DELETE envelope turns out to carry an empty `data: {}` rather than omitting the key, that path goes through the unmarshal branch instead — different code, same outcome for `ApiResponse`. Worth a check against a real endpoint before this lands.

kristina-solovyova (approved):

sergeyberezansky:
> Moved from the PR description, which is now release-notes shaped.

This PR adds proper pagination support for API responses that span multiple pages.

Previously, the `request` method had no way to signal to callers that the backend had truncated a response and more pages were available. The `NextToken` field in `ApiResponse` was also typed as `uuid.UUID`, which did not match the string tokens the backend actually returns.

The changes introduce an `ApiObjectResponse` interface with a `CombinePartialResponse` method. Response types that implement this interface opt into automatic pagination: `Request` will keep fetching subsequent pages using the `next_token` returned by the backend, folding each page into the caller's accumulated result before returning. Response types that do not implement the interface are fetched exactly once, preserving existing behavior.

`Quotas` is introduced as a named slice type that implements `ApiObjectResponse`, replacing the anonymous `[]Quota` used in `FindQuotaByFilter`. This was the immediate motivation: a filesystem can hold more quotas than the backend returns in a single response, so callers were silently receiving incomplete results.

Several bugs in the original pagination path are also fixed:

- The caller's `url.Values` is now cloned before `next_token` is set, so the caller's map is never mutated and a `nil` query no longer causes a panic on the second page.
- Each page is unmarshalled into a fresh object rather than reusing the caller's, preventing each page from overwriting the previous one.
- A hard limit of 1000 pages per request (`ApiMaxPagesPerRequest`) prevents an infinite loop if the backend keeps returning a token.
- `request` now returns the next token alongside the error, and retries reset the token so a failed attempt cannot leak its token into a subsequent retry.

Tests cover all four cases: multi-page accumulation, nil query handling, caller query immutability, and single-fetch behavior for non-paginated response types.

sergeyberezansky:
### Merge activity

* **Aug 10, 4:42 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/weka/csi-wekafs/715).
* **Aug 10, 4:56 PM UTC**: [Graphite](https://app.graphite.com/github/pr/weka/csi-wekafs/715) rebased this pull request as part of a merge.
* **Aug 10, 4:56 PM UTC**: @sergeyberezansky merged this pull request with [Graphite](https://app.graphite.com/github/pr/weka/csi-wekafs/715).

PR #717, opened 2026-08-08
feat: add TTL-based filesystem name cache to API client

sergeyberezansky:
### TL;DR
Filesystem lookups are now cached briefly, cutting repeated calls to the storage system.

### What changed?
- Filesystem lookups reuse a recent result for a short time instead of querying the storage system every time.
- Operations that need guaranteed fresh data are unaffected and always fetch live data.

### How to test?
1. This is an internal performance optimization, verified by automated tests included in this change.
2. Operators may notice fewer repeated filesystem lookup calls to the storage cluster when the same filesystems are used repeatedly.

### Why make this change?
Workloads that resolve many filesystems in quick succession were generating a burst of repeated, identical requests to the storage cluster.

PR #710, opened 2026-08-08
fix: optimize gc resource consumption and support tenants with same filesystem name 

sergeyberezansky:
### TL;DR
Fixes deleted-volume cleanup colliding across tenants sharing a filesystem name

### What changed?
- Cleanup of deleted volumes is now tracked per tenant, not just by filesystem name
- Fixes trash left behind indefinitely when two tenants both use a name like `default`
- Upgraded the deletion tool, lowering cleanup CPU/time cost

### How to test?
1. Create identically-named filesystems under two different tenants
2. Provision and delete volumes on each
3. Confirm both tenants' deleted data is fully cleaned up, without one blocking the other

### Why make this change?
A tenant's deleted volume data could previously be left behind indefinitely if another tenant used the same filesystem name — a real risk in shared, multi-tenant clusters.

PR #708, opened 2026-08-06
feat: implement external volume health monitoring via WEKA REST API

sergeyberezansky:
### TL;DR
Kubernetes can now detect and report when a volume's storage becomes unhealthy

### What changed?
- Periodic health checks for each volume against the Weka cluster
- Unhealthy volumes surface as abnormal conditions on the PVC
- Reports actual used capacity alongside health status
- On by default, checks every 5 minutes; requires Weka 4.3+

### How to test?
1. Deploy the CSI driver and create a PVC
2. Remove the volume's backing filesystem on the Weka cluster
3. Run `kubectl describe pvc <name>` and confirm an abnormal condition appears within 5 minutes

### Why make this change?
Previously Kubernetes couldn't tell if a volume's storage was still intact; problems went unnoticed until an application failed. This gives ongoing visibility into real volume health.

PR #718, opened 2026-08-08
fix: make ApiClient safe for concurrent use, and fix what review found

sergeyberezansky:
### TL;DR
Fixed bugs that could crash the driver, or make it misjudge storage cluster features, when many volume operations ran at once.

### What changed?
- Made the driver's storage API client safe for many volume operations running at once
- Fixed a partly-failed login being treated as successful, which left the driver with wrong assumptions about cluster capabilities for up to an hour (refusing valid volumes, skipping quota enforcement, or rejecting valid organizations)

### How to test?
Only reliably shown by automated tests, since the failures depend on timing under concurrent load:
1. Run the automated concurrency test suite added in this change
2. It reproduces the crashes and wrong behavior on the old code, and passes cleanly after the fix

### Why make this change?
Under concurrent load — normal in production, since Kubernetes creates and deletes many volumes in parallel — the driver could crash or silently run on wrong assumptions, causing hard-to-diagnose failures.

### DIFF
--- a/pkg/wekafs/apiclient/apiclient.go
+++ b/pkg/wekafs/apiclient/apiclient.go
@@ -51,12 +51,13 @@ type ApiClient struct {
 	CompatibilityMap           *WekaCompatibilityMap
 	clientHash                 uint32
 	hostname                   string
-	NfsInterfaceGroups         map[string]*InterfaceGroup
-	ApiUserRole                ApiUserRole
-	ApiOrgId                   int
-	containerName              string
-	NfsInterfaceGroupName      string
-	NfsClientGroupName         string
+	// nfsInterfaceGroups caches interface groups by name. It carries its own lock - see the type.
+	nfsInterfaceGroups    *interfaceGroups
+	ApiUserRole           ApiUserRole
+	ApiOrgId              int
+	containerName         string
+	NfsInterfaceGroupName string
+	NfsClientGroupName    string
 
 	containers           *ContainersResponse
 	containersUpdateTime time.Time
@@ -107,7 +108,7 @@ func NewApiClient(ctx context.Context, credentials Credentials, opts ApiClientOp
 		CompatibilityMap:   &WekaCompatibilityMap{},
 		hostname:           opts.Hostname,
 		apiEndpoints:       NewApiEndPoints(),
-		NfsInterfaceGroups: make(map[string]*InterfaceGroup),
+		nfsInterfaceGroups: newInterfaceGroups(),
 	}
 	a.resetDefaultEndpoints(ctx)
 	if len(a.Credentials.Endpoints) < 1 {
--- a/pkg/wekafs/apiclient/apiclient.go
+++ b/pkg/wekafs/apiclient/apiclient.go
@@ -122,8 +122,10 @@ func NewApiClient(ctx context.Context, credentials Credentials, opts ApiClientOp
 	return a, nil
 }
 
-// getBaseUrl returns the full HTTP URL of the API endpoint including schema, chosen endpoint and API prefix
-func (a *ApiClient) getBaseUrl(ctx context.Context) string {
+// getBaseUrl returns the full HTTP URL of the API endpoint including schema, chosen endpoint and API
+// prefix. It reports ApiNoEndpointsError, via requireEndpoint, rather than dereferencing a nil
+// endpoint when none is known.
+func (a *ApiClient) getBaseUrl(ctx context.Context) (string, apiError) {
 	scheme := ""
 	switch strings.ToUpper(a.Credentials.HttpScheme) {
 
@@ -134,8 +136,11 @@ func (a *ApiClient) getBaseUrl(ctx context.Context) string {
 	default:
 		scheme = "http"
 	}
-	endpoint := a.getEndpoint(ctx)
-	return fmt.Sprintf("%s://%s:%d/api/v2", scheme, endpoint.IpAddress, endpoint.MgmtPort)
+	endpoint, err := a.requireEndpoint(ctx)
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%s://%s:%d/api/v2", scheme, endpoint.IpAddress, endpoint.MgmtPort), nil
 }
 
 // handleTransientErrors checks if the error returned by endpoint is a network error (transient by definition)
@@ -174,9 +179,13 @@ func (a *ApiClient) handleTransientErrors(ctx context.Context, err error) error 
 }
 
 // getUrl returns a URL which consists of baseUrl + path
-func (a *ApiClient) getUrl(ctx context.Context, path string) string {
-	u, _ := url.JoinPath(a.getBaseUrl(ctx), path)
-	return u
+func (a *ApiClient) getUrl(ctx context.Context, path string) (string, apiError) {
+	base, err := a.getBaseUrl(ctx)
+	if err != nil {
+		return "", err
+	}
+	u, _ := url.JoinPath(base, path)
+	return u, nil
 }
 
 // generateHash used for storing multiple clients in hash table. Hash() is created once as connection params might change
--- a/pkg/wekafs/apiclient/apiclient.go
+++ b/pkg/wekafs/apiclient/apiclient.go
@@ -31,14 +31,17 @@ type ApiUserRole string
 // currentEndpointId: refers to the currently working API endpoint
 // Timeout sets max request timeout duration
 type ApiClient struct {
-	sync.Mutex
+	// Guards the mutable state below: Login writes it while concurrent requests read it.
+	sync.RWMutex
+	// loginMu serialises login attempts. Kept separate from RWMutex so a login, which makes several
+	// round-trips, never holds the state lock across I/O.
+	loginMu                    sync.Mutex
 	client                     *http.Client
 	Credentials                Credentials
 	ClusterGuid                uuid.UUID
 	ClusterName                string
 	MountEndpoints             []string
-	actualApiEndpoints         map[string]*ApiEndPoint
-	currentEndpoint            string
+	apiEndpoints               *ApiEndPoints
 	apiToken                   string
 	apiTokenExpiryDate         time.Time
 	refreshToken               string
@@ -93,7 +96,6 @@ func NewApiClient(ctx context.Context, credentials Credentials, opts ApiClientOp
 	}
 
 	a := &ApiClient{
-		Mutex: sync.Mutex{},
 		client: &http.Client{
 			Transport:     tr,
 			CheckRedirect: nil,
@@ -104,7 +106,7 @@ func NewApiClient(ctx context.Context, credentials Credentials, opts ApiClientOp
 		Credentials:        credentials,
 		CompatibilityMap:   &WekaCompatibilityMap{},
 		hostname:           opts.Hostname,
-		actualApiEndpoints: make(map[string]*ApiEndPoint),
+		apiEndpoints:       NewApiEndPoints(),
 		NfsInterfaceGroups: make(map[string]*InterfaceGroup),
 	}
 	a.resetDefaultEndpoints(ctx)
--- a/pkg/wekafs/apiclient/apiclient.go
+++ b/pkg/wekafs/apiclient/apiclient.go
@@ -35,15 +35,21 @@ type ApiClient struct {
 	sync.RWMutex
 	// loginMu serialises login attempts. Kept separate from RWMutex so a login, which makes several
 	// round-trips, never holds the state lock across I/O.
-	loginMu                    sync.Mutex
-	client                     *http.Client
-	Credentials                Credentials
-	ClusterGuid                uuid.UUID
-	ClusterName                string
-	MountEndpoints             []string
-	apiEndpoints               *ApiEndPoints
-	apiToken                   string
-	apiTokenExpiryDate         time.Time
+	loginMu            sync.Mutex
+	client             *http.Client
+	Credentials        Credentials
+	ClusterGuid        uuid.UUID
+	ClusterName        string
+	MountEndpoints     []string
+	apiEndpoints       *ApiEndPoints
+	apiToken           string
+	apiTokenExpiryDate time.Time
+	// loginComplete is true only once a login has published its token *and* finished the post-auth
+	// setup that follows (permission check, cluster info, endpoint discovery). Login clears it the
+	// moment it publishes a new token and sets it only on full success, so a goroutine that was
+	// blocked on loginMu while another goroutine's setup failed observes the failure - via
+	// loginSucceeded - instead of a token that merely exists.
+	loginComplete              bool
 	refreshToken               string
 	apiTokenExpiryInterval     int64
 	refreshTokenExpiryInterval int64
--- a/pkg/wekafs/apiclient/apiclient_race_test.go
+++ b/pkg/wekafs/apiclient/apiclient_race_test.go
@@ -1,0 +1,297 @@
+package apiclient
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strconv"
+	"sync"
+	"sync/atomic"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/rs/zerolog"
+)
+
+// quietLogs silences the client's per-request trace logging for the duration of a test. These tests
+// issue hundreds of requests, and what is under test is the race detector's verdict, not the output.
+// The level is restored afterwards so other tests in the package are unaffected.
+func quietLogs(t *testing.T) {
+	t.Helper()
+	previous := zerolog.GlobalLevel()
+	zerolog.SetGlobalLevel(zerolog.Disabled)
+	t.Cleanup(func() { zerolog.SetGlobalLevel(previous) })
+}
+
+// This file exercises ApiClient under `go test -race` to catch data races in its concurrency
+// primitives (the RWMutex-guarded auth state, the ApiEndPoints map, and per-endpoint counters).
+// It stands up a fake Weka REST API with httptest and never touches a real cluster or
+// localhost:14000.
+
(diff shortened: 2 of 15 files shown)


sergeyberezansky:
* **#741** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/741?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#740** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/740?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#733** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/733?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#726** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/726?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#725** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/725?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#724** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/724?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#723** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/723?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#721** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/721?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#751** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/751?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#750** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/750?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#720** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/720?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#719** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/719?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#718** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/718?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/718?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#717** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/717?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#716** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/716?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#715** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/715?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#714** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/714?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#713** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/713?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#712** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/712?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#711** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/711?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#710** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/710?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#709** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/709?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#708** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/708?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

kristina-solovyova:
@claude review

kristina-solovyova:
## Review notes — separating what this PR introduces from what it inherits

A review of `1169dd2c` turned up 10 findings. Most are not regressions from this PR, so sorting them by origin:

### Introduced here

- **`login.go:39` — the new post-`loginMu` re-check swallows a failed login.** `Login` publishes the token (L62-67) *before* `ensureSufficientPermissions`, `fetchClusterInfo`, and `UpdateApiEndpoints`. If one of those fails, `Login` returns an error but leaves `isLoggedIn()` true, so every goroutine that was blocked on `loginMu` hits the re-check and returns `nil` — success for a login that failed. Previously the waiter acquired `a.Lock()` with no re-check and performed a full login, which repaired the state. Concretely: with insufficient permissions, one caller correctly errors and all concurrent callers proceed to 403 on every request. This is the one behavioural regression in the change.
- **`apiendpoint.go:66` — `Replace`/`resetDefaultEndpoints` key namespaces disagree.** `Replace` tests membership with `endpoints[eps.currentEndpoint.String()]` (`ip:port`), while `resetDefaultEndpoints` still keys by the raw credential string (L169), which may carry no port. Before the change `currentEndpoint` *was* the raw key, so it matched. Consequences: the current selection is always dropped after a reset, and the default entries never match `existingEndpoints[endpointKey]` in `UpdateApiEndpoints`, so they're rebuilt as fresh objects (losing the stats this PR went to trouble to preserve). Related: the new warn loop (L228-232) means "Removing inactive API endpoint" now actually fires on every login — the old loop was dead code, as the description notes.
- **`apiendpoint.go:25` — `lastActive` is now write-only.** Its only reader went away with the prune loop, and the new comment ("when the endpoint was first discovered") is contradicted by `resetDefaultEndpoints`, which mints a new object with a new timestamp for the same endpoint. Worth dropping the field or restoring a use.
- **`apiclient_race_test.go:99` — `t.Fatalf` from HTTP handler goroutines.** `wrapData(t, …)` runs inside every `mux.HandleFunc`; `Fatalf` off the test goroutine calls `runtime.Goexit` and aborts the response mid-flight instead of failing the test, so a fixture marshalling error surfaces as an opaque client-side EOF and the test can still pass. `t.Errorf` + `http.Error` + return, or marshal the fixtures up front.

Two more are new code but not behaviour changes: `Current()` (L128) does check-then-act around `Rotate()` and can return nil under a concurrent `Replace` — the old `getEndpoint` had the same shape, racier; and `Rotate` samples uniformly including the currently selected endpoint, so the doc comment's "picks a *different* endpoint" isn't accurate (matters for the `handleTransientErrors` → `rotateEndpoint` failover path, which stays on a dead endpoint ~50% of the time with two endpoints).

### Pre-existing — not caused by this PR

- **`login.go:126` — a failed token refresh wedges the client into 401s (highest-severity finding overall).** When the refresh POST fails, `Init` falls back to `Login`, but `Login` returns `nil` immediately if `isLoggedIn()`, and `isLoggedInLocked` only tests `apiToken != ""` plus the *refresh* token's expiry — never `apiTokenExpiryDate`. So the request goes out with the expired access token, the 401 handler calls `Init` again, same no-op, and every request 401s until `refreshTokenExpiryDate` passes. Both the predicate and the fallback are semantically unchanged by this PR. (Note: the refresh path was already unserialised before — old `Init` took no lock at all — so this PR doesn't make the concurrent-refresh window any wider either.)
- **`interfacegroup.go:180` — `NfsInterfaceGroups` map is still unsynchronised.** Reads at L180/L187, write at L168, no lock of any kind. Reached concurrently from `nfsmount.go:156` and `nfs.go:855` on the shared per-cluster client, so two simultaneous NFS publishes on a cold cache give `fatal error: concurrent map read and map write`. The file isn't in this diff — but it is the same failure mode the PR exists to eliminate, so it's a reasonable candidate to fold in.
- **`requests.go:65` — stats can be attributed to the wrong endpoint.** `do()` resolves the endpoint twice, once inside `getUrl` (L27) and again for the counters (L65); a rotation in between lands the counts on an endpoint that was never contacted. Both call sites predate this PR, though it does blunt the value of the new atomics.

### Pre-existing but only half-addressed here

- **One-sided locking on exported cluster state.** `cluster.go` now writes `ApiOrgId`, `ClusterName`, and `ClusterGuid` under `a.Lock()`, but the out-of-package readers weren't converted: `nfsmount.go:174,176`, `volume.go:458`, `apistore.go:65,187`. A write lock doesn't synchronise an unlocked reader, so these are still races under the Go memory model and `ClusterGuid` can be read torn. Not a regression, but it sits inside what "safe for concurrent use" promises — either add accessors or keep those fields immutable after first login.

---
🤖 Posted by [Claude Code](https://claude.com/claude-code)

sergeyberezansky:
Fixed in four commits, one per concern — `1a3e5ae4`, `31f887fd`, `879252b0`, `77006f1a`. Going through the findings in the order you sorted them, since that framing was the useful part.

## Introduced here

**The re-check swallowing a failed login** — fixed in `77006f1a`. Kept publishing the token early: the post-auth calls go through `Get`/`Request` and need it to authenticate, so publishing after setup would break the very calls that complete the setup. Added `loginComplete`, written in the same critical section that publishes the token — cleared when the token goes out, set only once `ensureSufficientPermissions` and `fetchClusterInfo` have succeeded. The re-check now asks whether the last login *completed*, not whether a token exists. (`UpdateApiEndpoints` failing is already non-fatal, so it does not hold the flag back.)

**`Replace`/`resetDefaultEndpoints` key mismatch** — fixed in `1a3e5ae4`. Both sides key on `String()` now. Your read was right that this quietly defeats the PR's own goal: the default entries were being rebuilt as fresh objects on every reset, discarding the counters the atomics exist to accumulate.

**`lastActive` write-only** — dropped, with its writes and the now-unused `time` import.

**`t.Fatalf` from handler goroutines** — fixed in `77006f1a`. `wrapData` returns an error; a new `writeJSON` does `t.Errorf` + 500 + return. Worth noting this bit for real during the work: the two tests I added on top of it were themselves broken by a later commit in the stack, and neither failed at this PR — see below.

**`Current()` check-then-act / `Rotate` sampling the current endpoint** — `Rotate` now excludes the current endpoint when more than one is known, and returns what it picked so `rotateEndpoint` stops re-reading the selection (which could log an endpoint the call never chose).

## Half-addressed here, now finished

**One-sided locking on exported cluster state** — fixed in `31f887fd`. Added `GetClusterName`/`GetClusterGuid`/`GetApiOrgId` under `RLock` and converted `nfsmount.go:174,176`, `volume.go:458`, `apistore.go:65,187`, plus the unlocked reads in this PR's own race test. Exactly the treatment `CompatibilityMap`/`ApiUserRole` already had — these were simply missed.

## Pre-existing, folded in

**`NfsInterfaceGroups`** — fixed in `879252b0`, and reworked after review feedback into an `interfaceGroups` type owning its `RWMutex` (the shape `ApiEndPoints` already uses) rather than a bare map plus a sibling mutex on `ApiClient`. The field is unexported now, which is the actual fix: an exported map invites precisely the unlocked access that caused this. Its lock is deliberately separate from the client's, because the cache is filled through `Request`, which takes the client lock — sharing it would deadlock. Verified by removing the locking and re-running under `-race`, which reports the race the type exists to prevent.

## Deliberately not fixed

**The refresh-failure wedge into permanent 401s.** Agreed it is the highest-severity finding overall, and agreed it is untouched by this PR — both the predicate and the fallback are unchanged. Changing `isLoggedInLocked` to consult `apiTokenExpiryDate` alters auth semantics well beyond a concurrency fix, so it is tracked separately rather than smuggled in here.

**`requests.go:65` double endpoint resolution** — left as is. Real, pre-existing, and only cosmetic in effect (counters and error text can name an endpoint that never served the request).

## Two things the split turned up

The commits are one-per-concern, and each builds on its own — I checked branch by branch rather than only at the tip. Two of them could not be separated: findings 1 and 4 share `77006f1a` because `login_test.go` uses the harness helper, so splitting them yields a commit that does not compile.

More interestingly, **both tests I added here were broken by #720**, which replaces this file's harness with the in-process fake and deletes `newRaceTestServer` and `writeJSON`. Neither failure appeared at this PR — only two commits later. Fixed at both ends: `login_test.go` now defines its own JSON helper so a commit that never touches it cannot break it, and the interface-group race test is ported onto `apiclienttest` in #720 by its own commit. That is the same failure mode as the `t.Fatalf` finding, one level up: a test that silently stops being a test.

## Testing

Each fix has a test that fails when that fix alone is reverted — verified for the login gate, the endpoint key mismatch, the interface-group lock, `Rotate`, and the nil-endpoint path.

Note this PR is **not independently green on unit tests**: `TestVolumeId` and `TestMountOptions_AsNfs` fail at this commit and are fixed higher in the stack (#721 and #719). Confirmed pre-existing by building a detached worktree at the pre-fix commit, so it is not something these changes introduced. `TestGetNfsInterfaceGroup`, `TestApiClient_ResolvePathToInode` and `TestApiClientRequest` need a live cluster at `localhost:14000`.

kristina-solovyova (approved):

sergeyberezansky:
> Moved from the PR description, which is now release-notes shaped.

Makes `ApiClient` safe for concurrent use: `RWMutex` in place of `Mutex`, atomic per-endpoint counters, and endpoint selection moved behind `ApiEndPoints`.

Four follow-up commits address the review. Sorted by origin, because most of what turned up was not introduced here.

## Introduced by this PR

**`1a3e5ae4` — endpoint selection.** Four defects, three of them new:

- `Replace` and `resetDefaultEndpoints` disagreed on the map key (`ip:port` vs the raw credential string, which may carry no port). Before the rework `currentEndpoint` *was* the raw key, so they matched. The current selection was therefore dropped after every reset, and default entries never matched in `UpdateApiEndpoints` — so they were rebuilt as fresh objects, discarding the per-endpoint counters this PR exists to preserve.
- `Rotate` documented picking a *different* endpoint but sampled uniformly including the current one. `request()` calls it after a transient error specifically to move off a failing node; with two endpoints the retry stayed on the dead one about half the time. It now excludes the current endpoint and returns what it picked, so `rotateEndpoint` stops re-reading the selection and logging an endpoint it never chose.
- `getEndpoint` could return nil and every caller dereferenced it. `NewApiClient` only rejects an *empty* endpoint list, so a secret whose entries all fail validation leaves an empty map and the first call panics. Returns `ApiNoEndpointsError` now.
- `lastActive` had become write-only; removed.

**`6293401f` — a failed login reported as success.** The token is published before the post-auth setup (`ensureSufficientPermissions`, `fetchClusterInfo`, `UpdateApiEndpoints`), because those calls go through `Request` and need it. The re-check added after `loginMu` tested only `isLoggedIn()`, which the published token already satisfies — so when setup failed, the goroutine doing the login returned an error while every waiter returned `nil`.

Those waiters then ran with a zero-value `CompatibilityMap`: filesystem- and snapshot-backed `CreateVolume` refused as unsupported, quota enforcement silently degrading to xattr so PVCs get no enforced limit, non-Root organizations rejected with a spurious upgrade message — for the ~1h life of the access token. Before this PR the waiter had no re-check and simply re-ran the login, which repaired the state. Now tracked by `loginComplete`, written in the same critical section that publishes the token.

**`6293401f` — race-test harness.** `wrapData` ran on the handler goroutine and called `t.Fatalf`, which invokes `runtime.Goexit`: the response is abandoned mid-flight, the client sees an opaque EOF, and **the test can still pass**. Now returns an error, with `writeJSON` turning it into `t.Errorf` plus a 500.

## Half-addressed by this PR, now finished

**`31f887fd` — cluster identity.** `ApiOrgId`, `ClusterName` and `ClusterGuid` were given write-side locking here, but every reader outside the package was left unlocked (`nfsmount.go`, `apistore.go`, `volume.go`). A write lock does not synchronise an unlocked reader, and `ClusterGuid` is wide enough to tear. Added `GetClusterName`/`GetClusterGuid`/`GetApiOrgId` under `RLock` and converted the call sites — the same treatment `CompatibilityMap` and `ApiUserRole` already had.

## Pre-existing, folded in because it is the same failure class

**`879252b0` — NFS interface group cache.** An exported plain map with no synchronisation at all, reached concurrently from `nfsmount.go` and `nfs.go` via the shared per-cluster client. Two NFS publishes on a cold cache give `fatal error: concurrent map read and map write` — a crash, not a detectable race. Now an `interfaceGroups` type owning its own `RWMutex` (the shape `ApiEndPoints` uses) with the field unexported, so the map cannot be reached unguarded. Its lock is deliberately separate from the client's: the cache is filled through `Request`, which takes the client lock, so sharing it would deadlock.

## Deliberately left out

`isLoggedInLocked` still tests only `apiToken != ""` and the *refresh* token's expiry, never `apiTokenExpiryDate`. When a refresh fails, `Init` falls back to `Login`, which returns early because `isLoggedIn()` is still true, and every request 401s until the refresh window closes. It is the highest-severity finding overall, but it predates this PR, is unchanged by it, and altering that predicate has wider auth consequences than this change should carry. Tracked separately.

## Testing

Each fix has a test that fails when that fix alone is reverted; verified for the login gate, the endpoint key mismatch, the interface-group lock (reproduces `DATA RACE` when removed), `Rotate`, and the nil-endpoint path. Every commit builds on its own.

Note `TestGetNfsInterfaceGroup`, `TestApiClient_ResolvePathToInode` and `TestApiClientRequest` fail without a live cluster at `localhost:14000` — pre-existing and unrelated. `TestVolumeId` and `TestMountOptions_AsNfs` also fail at this commit; they are fixed higher in the stack (#721 and #719 respectively), so this PR is not independently green on unit tests.

sergeyberezansky:
### Merge activity

* **Aug 10, 4:42 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/weka/csi-wekafs/718).
* **Aug 10, 5:02 PM UTC**: [Graphite](https://app.graphite.com/github/pr/weka/csi-wekafs/718) rebased this pull request as part of a merge.
* **Aug 10, 5:02 PM UTC**: @sergeyberezansky merged this pull request with [Graphite](https://app.graphite.com/github/pr/weka/csi-wekafs/718).

PR #743, opened 2026-08-09
feat: label volume metrics with the Weka organization

sergeyberezansky:
### TL;DR
Added a Weka organization (tenant) label to per-volume Prometheus metrics.

### What changed?
- Per-volume metrics now carry an `organization` label, taken from the credentials the API client authenticated with
- Root-organization credentials are labeled `Root` rather than exported blank
- Added tests validating the label list and values stay in sync and correctly ordered

### How to test?
Covered by automated tests only.

### Why make this change?
Volume capacity and usage metrics could previously only be grouped by volume or filesystem. Since every volume belongs to exactly one organization, adding the label makes existing series groupable and rankable per tenant at no cost in additional series.

---

> [!NOTE]
> **Low Risk**
> Test-only addition with no runtime or production code changes in this diff.
> 
> **Overview**
> Adds **`organization_label_test.go`** with automated coverage for the Weka **organization** tenant label on per-volume Prometheus metrics.
> 
> Tests lock in **`organizationLabel`** behavior (named tenant, explicit `Root`, empty credentials, nil client → root), assert **`createPrometheusLabelsForMetric`** returns values that match **`LabelsForCsiVolumes`** in count and position (including `organization`, `filesystem_name`, `pv_name`), and verify the built slice is accepted by a real **`GaugeVec`** using **`LabelsForCsiVolumes`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63ca63b505449dc80bd1273ea9b24cf9d90477db. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #742, opened 2026-08-09
feat: fetch quotas in filesystem batches by default

sergeyberezansky:
### TL;DR
Made batched, per-filesystem quota fetching the metrics server's default.

### What changed?
- `enableBatchModeForQuotaUpdates` now defaults to `true` in both the `csi-metricsserver` and `csi-wekafsplugin` charts
- The `--fetchquotasinbatchmode` flag is now always rendered with its value, so it can still be set to `false`

### How to test?
1. Deploy the metrics server without setting `enableBatchModeForQuotaUpdates`.
2. Confirm quota requests go to the WEKA API per filesystem rather than per volume.
3. Set the value to `false` and confirm it falls back to one request per volume.

### Why make this change?
Directory-backed volumes — the common case — share a filesystem, so fetching quotas per filesystem instead of per volume cuts WEKA API load sharply (measured ~33 requests/second down to 5 on a 10k-volume fixture) with no downside for filesystem-backed volumes. The trade-off, unchanged from before, is that reported values can be as stale as `quotaCacheValiditySeconds`.

---

> [!NOTE]
> **Medium Risk**
> Changing the default alters WEKA API traffic and metric freshness for existing deployments that did not set the value explicitly; behavior remains reversible via Helm or the flag.
> 
> **Overview**
> **Batched per-filesystem quota fetching is now the default** for the metrics server in both Helm charts and the `--fetchquotasinbatchmode` CLI flag (`metricsserver` and `wekafsplugin`).
> 
> Helm deployments **always pass** `--fetchquotasinbatchmode={{ .Values.metricsServer.enableBatchModeForQuotaUpdates }}` instead of only adding the flag when enabled, so operators can **explicitly turn batch mode off** even though the binary default is now `true`. Chart default `enableBatchModeForQuotaUpdates` changes from `false` to `true`, with values comments rewritten to describe the API-load vs. cache-freshness trade-off.
> 
> **Operational effect:** new installs and upgrades that do not override the value should see far fewer WEKA API quota calls (one per filesystem rather than per volume), with metrics potentially up to `quotaCacheValiditySeconds` stale unless `honorTimestamps` is in use.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 61bf8c0a5d17eabcaaa5f0fd6ebf20e4e50b3f74. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #740, opened 2026-08-09
fix: drop the metrics server readinessProbe, keep /readyz

sergeyberezansky:
### TL;DR
Removed the metrics server's readiness probe so standby pods stop reporting NotReady forever.

### What changed?
- Dropped the `readinessProbe` (which checked `/readyz`, a leadership indicator) from the metrics server Deployment template in both `csi-metricsserver` and `csi-wekafsplugin` charts
- `/readyz` itself is still served, just no longer gated by a probe
- Removed `charts/csi-metricsserver/ci/ct-values.yaml`, which only existed to pin chart-testing to one replica

### How to test?
1. Deploy the metrics server with more than one replica.
2. Confirm non-leader pods show `Ready` in `kubectl get pods` instead of staying `NotReady` indefinitely.

### Why make this change?
`/readyz` reports leadership, not health, so every non-leader replica was permanently NotReady. Nothing routes traffic based on that readiness state, but on clusters running kube-prometheus it permanently triggered `KubePodNotReady` and `KubeDeploymentReplicasMismatch` alerts for a healthy deployment.

---

> [!NOTE]
> **Low Risk**
> Helm-only probe and CI values changes; no application binary changes, with slightly looser readiness semantics that match current lack of a Service fronting these pods.
> 
> **Overview**
> **Removes the metrics server Kubernetes `readinessProbe`** (HTTP `/readyz` on the health port) from the `csi-metricsserver` and embedded `csi-wekafsplugin` deployment templates. **Liveness on `/healthz` is unchanged**; `/readyz` remains available on the process but is no longer used for pod readiness.
> 
> With leader election, `/readyz` reflects **leadership**, so every standby stayed **NotReady** indefinitely. That did not affect restarts (only liveness does) but caused persistent **`KubePodNotReady`** and **`KubeDeploymentReplicasMismatch`** on kube-prometheus clusters even when the deployment was healthy. The templates now document that readiness should return only if a **Service** needs endpoint gating.
> 
> **Deletes `charts/csi-metricsserver/ci/ct-values.yaml`**, which had forced a single replica so chart-testing’s `--wait` would not hang on standby NotReady pods.
> 
> **Values comments** in both charts now state that standby replicas show **Ready** and leadership is visible in the **Lease**, not pod status.
> 
[...]

PR #739, opened 2026-08-09
fix: give the metrics server its own leader election lease

sergeyberezansky:
### TL;DR
Gave the metrics server its own leader-election lease so it can't stall the CSI controller.

### What changed?
- The metrics server now uses a separate leader-election lease name from the CSI controller
- The controller's lease name is unchanged, to avoid split leadership across a rolling upgrade

### How to test?
1. Enable `metricsServer.enabled` on the `csi-wekafsplugin` chart alongside the controller.
2. Run `kubectl get lease` and confirm the controller and metrics server each hold a distinct lease.
3. Confirm the controller acquires leadership, its gRPC server starts, and new PVCs provision normally.

### Why make this change?
Both processes previously competed for one lease. On a live cluster, the metrics server won it: both controller pods sat waiting for leadership, the gRPC server never started, volume health reporting stopped, and new PVCs stayed Pending indefinitely — while every pod still reported Running, so nothing looked wrong from the outside.

---

> [!NOTE]
> **Medium Risk**
> Touches CSI controller startup and leader election—critical for provisioning—but the change is narrow (lease naming only) with explicit tests and an unchanged controller lease ID for upgrades.
> 
> **Overview**
> Fixes a deadlock where the in-chart metrics server and CSI controller both competed for the same Kubernetes leader-election lease (`{driver}-controller-leader`). When metrics pods won that lease, controller replicas never became leader—gRPC never started, sidecars stayed on wait-for-leader, and provisioning stalled while pods looked healthy.
> 
> **`leaderElectionIDForMode`** now assigns **`{driver}-metricsserver-leader`** only for `CsiModeMetricsServer`; controller, `CsiModeAll`, and other modes still use the unchanged **`{driver}-controller-leader`** so rolling upgrades do not split controller leadership. `initManager` sets `LeaderElectionID` via this helper instead of a single hardcoded name.
> 
> Unit tests lock in: metrics vs controller leases differ, `CsiModeAll` matches controller, the controller lease string is unchanged, and lease names vary by driver name.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1597454ada735d8e915a98fc6f16ba6a829eda19. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #764, opened 2026-08-11
fix: make --allowprotocolcontainers actually reach the mount

sergeyberezansky:
### TL;DR
Four unrelated defects fixed together; the operator-visible one is that `--allowprotocolcontainers` had no effect at all.

### What changed?
- `--allowprotocolcontainers` now reaches the mount. The flag was accepted, and reported as enabled at startup, but every mount behaved as though it were off.
- A snapshot that fails to delete now reports the snapshot's name. The error named the filesystem instead, so a failing snapshot deletion read as a filesystem problem.
- `MAXIMUM_PERSISTENT_VOLUME_COUNT` is validated. A value that is unparseable, zero, negative or above 2147483647 is now logged and ignored so the default limit stands, instead of being applied as given.
- A retried deletion of the same volume no longer logs a failure for work another attempt had already finished (CSI-380). The deletion itself reported success either way.

### How to test?
1. Add `--allowprotocolcontainers` to the node plugin's arguments — it is a container argument on the DaemonSet, with no Helm value of its own — on a cluster whose clients are protocol containers. Mount a PVC and confirm no `Failed to ensure local container` error is logged. Without the flag reaching the mount, that error appeared no matter how the flag was set.
2. Set `MAXIMUM_PERSISTENT_VOLUME_COUNT` on the metrics server to `abc`, then to `-1`. Each should be logged as rejected, with the default limit left in place.
3. Delete a VolumeSnapshot whose snapshot cannot be removed, and confirm the returned error names the snapshot rather than its filesystem.
4. The retried-deletion fix only shows up under concurrency, so it is covered by the unit tests rather than by anything you can do by hand: `go test ./pkg/wekafs/`.

### Why make this change?
Each of these failed quietly instead of loudly, which is the expensive kind. A flag that is accepted, logged as enabled and then ignored is worse than one that is rejected outright. An error naming the wrong object sends whoever reads it to the wrong place. An out-of-range limit silently shrinks how many PersistentVolumes the metrics server will look at. And a retried delete logged a failure for work that had in fact succeeded.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Mount and volume-delete paths change behavior operators rely on; GC locking is new but scoped to local rename only. Metrics and error-message fixes are low blast radius.
> 
> **Overview**
> Fixes four quiet failures: **`--allowprotocolcontainers`** is now read from driver config at mount time in `ensureLocalContainerName` instead of an unused field on the mounter/mount, so protocol-container clients can mount.
> 
> **Volume delete / CSI-380:** trash moves use a dedicated `trashMu` and new `renameIntoTrash` so concurrent or retried `DeleteVolume` calls cannot both rename the same path; the loser treats an already-moved source as success. A concurrent unit test covers this.
> 
> **Metrics server:** `MAXIMUM_PERSISTENT_VOLUME_COUNT` that is non-numeric, ≤0, or above `MaxInt32` is warned and ignored; the default limit stays.
> 
> **Snapshot delete:** internal errors now reference the **snapshot** name, not the filesystem.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c2a65c4c592ca12b9bd86244f233cb2d3a39a27. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/wekafs/gc.go
+++ b/pkg/wekafs/gc.go
@@ -43,6 +43,10 @@ type innerPathVolGc struct {
 	isRunning  map[gcKey]bool
 	isDeferred map[gcKey]bool
 	sync.Mutex
+	// trashMu serialises the check-then-rename in moveVolumeToTrash. It is deliberately not the
+	// embedded Mutex above, which initiateGarbageCollection holds - moveVolumeToTrash defers a
+	// call to it, so sharing the lock would make the two deadlock-prone by ordering alone.
+	trashMu sync.Mutex
 	mounter AnyMounter
 	config  *DriverConfig
 }
@@ -86,22 +90,9 @@ func (gc *innerPathVolGc) moveVolumeToTrash(ctx context.Context, volume *Volume)
 		return err
 	}
 	volumeTrashLoc := filepath.Join(path, garbagePath)
-	if err := os.MkdirAll(volumeTrashLoc, DefaultVolumePermissions); err != nil {
-		if !os.IsExist(err) {
-			logger.Error().Str("garbage_collection_path", volumeTrashLoc).Err(err).Msg("Failed to create garbage collector directory")
-			return err
-		}
-	}
 	fullPath := filepath.Join(path, volume.GetFullPath(ctx))
 	logger.Debug().Str("full_path", fullPath).Str("volume_trash_location", volumeTrashLoc).Msg("Moving volume contents to trash")
-	newPath := filepath.Join(volumeTrashLoc, filepath.Base(fullPath))
-	if !fileExists(fullPath) {
-		logger.Debug().Str("full_path", fullPath).Msg("Volume contents not found, maybe already moved to trash, skipping")
-		return nil
-	}
-	if err := os.Rename(fullPath, newPath); err != nil {
-		logger.Error().Err(err).Str("full_path", fullPath).
-			Str("volume_trash_location", volumeTrashLoc).Msg("Failed to move volume contents to volumeTrashLoc")
+	if err := gc.renameIntoTrash(ctx, fullPath, volumeTrashLoc); err != nil {
 		return err
 	}
 	// NOTE: there is a problem of directory leaks here. If the volume innerPath is deeper than /csi-volumes/vol-name,
@@ -112,6 +103,39 @@ func (gc *innerPathVolGc) moveVolumeToTrash(ctx context.Context, volume *Volume)
 	// 2024-07-29: apparently seems this is not a real problem since static volumes are not deleted this way
 	//             and dynamic volumes are always created inside the /csi-volumes
 	logger.Debug().Str("full_path", fullPath).Str("volume_trash_location", volumeTrashLoc).Msg("Volume contents moved to trash")
+	return nil
+}
+
+// renameIntoTrash moves fullPath into volumeTrashLoc, creating the trash directory if needed, and
+// does nothing when the source is already gone.
+//
+// The whole sequence is serialised: two concurrent attempts on the same volume - a retried
+// DeleteVolume being the ordinary case - could both observe the source present, and the loser would
+// then fail renaming a path the winner had already moved, leaving the volume stuck in Deleting
+// (CSI-380). The critical section covers only these local filesystem calls, never the mount that
+// precedes them, so it costs nothing measurable on the delete path.
+func (gc *innerPathVolGc) renameIntoTrash(ctx context.Context, fullPath, volumeTrashLoc string) error {
+	logger := log.Ctx(ctx)
+
+	gc.trashMu.Lock()
+	defer gc.trashMu.Unlock()
+
+	if err := os.MkdirAll(volumeTrashLoc, DefaultVolumePermissions); err != nil {
+		if !os.IsExist(err) {
+			logger.Error().Str("garbage_collection_path", volumeTrashLoc).Err(err).Msg("Failed to create garbage collector directory")
+			return err
+		}
+	}
+	if !fileExists(fullPath) {
+		logger.Debug().Str("full_path", fullPath).Msg("Volume contents not found, maybe already moved to trash, skipping")
+		return nil
+	}
+	newPath := filepath.Join(volumeTrashLoc, filepath.Base(fullPath))
+	if err := os.Rename(fullPath, newPath); err != nil {
+		logger.Error().Err(err).Str("full_path", fullPath).
+			Str("volume_trash_location", volumeTrashLoc).Msg("Failed to move volume contents to volumeTrashLoc")
+		return err
+	}
 	return nil
 }
 
--- a/pkg/wekafs/gc_test.go
+++ b/pkg/wekafs/gc_test.go
@@ -2,6 +2,9 @@ package wekafs
 
 import (
 	"context"
+	"os"
+	"path/filepath"
+	"sync"
 	"testing"
 
 	"github.com/wekafs/csi-wekafs/pkg/wekafs/apiclient"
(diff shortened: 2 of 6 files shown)


sergeyberezansky:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/764?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#766** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/766?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#767](https://github.com/weka/csi-wekafs/pull/767) <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/767?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#777](https://github.com/weka/csi-wekafs/pull/777) <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/777?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#765** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/765?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#764** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/764?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/764?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#740** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/740?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#741](https://github.com/weka/csi-wekafs/pull/741) <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/741?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#725** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/725?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#726](https://github.com/weka/csi-wekafs/pull/726) <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/726?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#723** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/723?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#724](https://github.com/weka/csi-wekafs/pull/724) <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/724?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#826** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/826?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#803** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/803?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#802** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/802?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#801** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/801?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#800** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/800?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#799** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/799?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#798** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/798?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#797** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/797?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#796** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/796?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#795** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/795?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#791** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/791?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#786** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/786?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#701** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/701?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#787** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/787?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`


<h2></h2>

sergeyberezansky:
The description is now in the house format and covers all four fixes rather than only the one in
the title. Keeping them in one PR deliberately - the stack has no capacity to split them.

One claim was corrected rather than reworded, so it is worth recording why.

The previous text said of the CSI-380 fix: *"The error propagated out of DeleteVolume, so the volume
stayed in Deleting and the next retry hit the same window."* That is not what happened. The losing
attempt's rename fails with `ENOENT`, which arrives as an `*os.LinkError`, and `os.IsNotExist` on it
is true - I checked by running it rather than reading it. `DeleteVolume` tests `os.IsNotExist` ahead
of its error branch and returns success for idempotence, and nothing wraps the error on the way up.
So the observable symptom was a spurious `ERROR`-level log line, not a volume stuck in `Deleting`.

The fix is still worth having - the sequence was nondeterministic and logged a failure for work that
had succeeded - but if the CSI-380 ticket reports volumes genuinely stuck in `Deleting`, this is not
its cause, and closing the ticket on this PR would hide whatever is.

Two smaller notes on the old text, now dropped from the description:

- It said the ineffective flag had "nothing logged or warned about it". The startup config dump does
  print `allow_protocol_containers`, so an operator saw it reported as enabled while it did nothing.
  That is worse than silence, and the new description says so.
- One commit subject says the volume-count override was "narrowing it blindly". On a 64-bit build
  `int` is 64 bits, so nothing was being truncated; rejecting values above 2147483647 is a new
  conservative bound rather than a truncation fix.

Also unmentioned by the old text and left as-is for now: after the trash move was extracted, the
"Volume contents moved to trash" debug line is emitted even when the contents were already gone, so
it now claims a move that did not happen. Cosmetic, and I can fold it in if you want it gone.

<details>
<summary>The previous description, verbatim</summary>

fix: make --allowprotocolcontainers actually reach the mount

The driver config carried the flag, but wekafsMounter and wekafsMount each
declared their own allowProtocolContainers field and newWekafsMounter never
assigned it. The always-false mounter field was copied into every mount and
handed to EnsureLocalContainer, so setting the flag had no effect at all and
nothing logged or warned about it.

Read it from the driver config at the point of use, next to wekafsContainerName
which was already sourced that way, and delete both intermediate fields. The bug
was that a value could be copied down two levels and silently miss a step; with
the fields gone there is no copy left to go stale.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

fix(CSI-380): serialise the GC trash move so a retried delete cannot get stuck

Moving a volume's contents to the trash directory is a check-then-rename, and it
ran unsynchronised. Two attempts on the same volume - which is exactly what a
retried DeleteVolume produces - could both observe the source present, and
whichever lost the race then failed renaming a path the other had already moved.
The error propagated out of DeleteVolume, so the volume stayed in Deleting and
the next retry hit the same window.

Extract the sequence as renameIntoTrash and take a lock across it. The lock is a
dedicated mutex rather than the collector's existing one, which
initiateGarbageCollection holds and moveVolumeToTrash defers a call to - sharing
it would leave the two safe only by defer ordering. The critical section covers
the local filesystem calls only, never the mount that precedes them.

The test runs 16 concurrent attempts at one volume and asserts every one reports
success with the contents moved exactly once. Without the lock it fails on
essentially every run.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

fix: report the snapshot name when snapshot deletion fails

deleteSnapshot returned "Failed to delete filesystem <filesystem>" on the error
path, naming the wrong object and the wrong field. The message reaches Kubernetes
as the DeleteSnapshot gRPC error, so a failing snapshot deletion read as a
filesystem problem.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

fix: validate MAXIMUM_PERSISTENT_VOLUME_COUNT instead of narrowing it blindly

The override was parsed as a 64-bit integer and converted straight to int with no
range check and no message when it failed to parse, so a typo silently kept the
default while a negative or out-of-range value became a nonsensical limit on the
PersistentVolume list.

Reject anything unparseable, non-positive or wider than an int32, log which it
was, and keep the default.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

</details>

Issue #691, opened 2026-06-09
Feature Request: Separate `priorityClassName` for Controller and Node components


Alja9:
Hi Team,

**Current behavior**
The `priorityClassName` value is a single global field applied to both the controller `Deployment` and node `DaemonSet`.
https://github.com/weka/csi-wekafs/blob/6dd6f79787b41a86995d25c186a6bbd7e39318d6/charts/csi-wekafsplugin/values.yaml#L170-L171

**Expected behavior**
Allow setting `priorityClassName` per component under `controller.priorityClassName` and `node.priorityClassName`, similar to how `nodeSelector` and `affinity` are already handled in this chart.

This is needed because the two components have different scheduling requirements, example:
- Controller → `system-cluster-critical`
- Node → `system-node-critical`

**Proposed change**
Add `controller.priorityClassName` and `node.priorityClassName` in `values.yaml`, and if possible using `coalesce` in the templates so the global `priorityClassName` still works as a fallback - fully backward compatible.

Thanks !

PR #778, opened 2026-08-11
feat: allow a separate priorityClassName for controller and node components

sergeyberezansky:
### TL;DR
The controller and the node pods can now be given different priority classes.

### What changed?
- New chart values `controller.priorityClassName` and `node.priorityClassName`.
- Either one overrides the existing global `priorityClassName` for its own pods. Left unset, both inherit it, so nothing changes for an existing installation.

### How to test?
1. Install with `--set controller.priorityClassName=system-cluster-critical --set node.priorityClassName=system-node-critical`.
2. Confirm `kubectl get deploy,ds -n csi-wekafs -o custom-columns=NAME:.metadata.name,PC:.spec.template.spec.priorityClassName` shows the two different classes.
3. Install with only the global `priorityClassName` set and confirm both components still use it.

### Why make this change?
Requested in issue #691. A controller Deployment and a node DaemonSet have different scheduling requirements — the usual pairing is `system-cluster-critical` for the controller and `system-node-critical` for the node — and a single global value could only ever express one of them. Setting a per-component value replaces the global one rather than being combined with it, since a priority class is a single name.


sergeyberezansky:
* **#778** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/778?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/778?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

Issue #687, opened 2026-06-04
The `csi-snapshotter` container cannot use custom `logLevel`


Alja9:
Hi Team,

The `csi-snapshotter` container cannot use custom `logLevel` from `values.yaml` :
https://github.com/weka/csi-wekafs/blob/6dd6f79787b41a86995d25c186a6bbd7e39318d6/charts/csi-wekafsplugin/values.yaml#L164-L165

Because the template uses hardcoded `v=5` :
https://github.com/weka/csi-wekafs/blob/5a08689096e87cf261e440a4cb8d7fbdbaebc127/charts/csi-wekafsplugin/templates/controllerserver-deployment.yaml#L351-L352

It would be better if there is no hardcoded in the template, thanks !

PR #779, opened 2026-08-11
fix(chart): use the configured logLevel for the csi-snapshotter sidecar

sergeyberezansky:
### TL;DR
`logLevel` now applies to the snapshotter sidecar too, instead of it always logging at level 5.

### What changed?
- The `csi-snapshotter` container's verbosity came from a value written into the template rather than from `logLevel`. It now follows `logLevel` like every other container.
- No change at the default: the hardcoded value and the default are both 5.

### How to test?
1. Install with `--set logLevel=2`.
2. Run `kubectl get deploy <release>-controller -o yaml` and confirm the `csi-snapshotter` container is started with `--v=2`.
3. Confirm its logs are correspondingly quieter.

### Why make this change?
Reported in issue #687. Turning `logLevel` down quieted every container except the snapshotter, which kept logging at 5 — and on a busy controller that was most of the remaining log volume. While fixing it we checked every other container in both charts; this was the only one whose log level was not configurable.

### DIFF
--- a/charts/csi-wekafsplugin/templates/controllerserver-deployment.yaml
+++ b/charts/csi-wekafsplugin/templates/controllerserver-deployment.yaml
@@ -353,7 +353,7 @@ spec:
           command: ["/shared/wait-for-leader"]
           args:
             - "/csi-snapshotter"
-            - "--v=5"
+            - "--v={{ .Values.logLevel | default 5 }}"
             - "--csi-address=$(ADDRESS)"
             - "--timeout=60s"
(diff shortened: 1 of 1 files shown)


sergeyberezansky:
* **#779** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/779?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/779?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

PR #793, opened 2026-08-13
fix: stop the driver panicking when it cannot determine volume encryption

sergeyberezansky:
### TL;DR

Fixes a crash: the driver could panic when checking whether a volume is encrypted, taking the process down.

### What changed?

- Checking encryption on a filesystem-backed volume with no API credentials bound now returns a clear error instead of crashing.
- The error says the encryption state could not be determined, and includes the underlying reason when there is one.
- Volume creation that needs this answer fails with an `Internal` error rather than continuing.

### How to test?

1. Create a volume backed by a whole filesystem whose storage class has no API secret attached.
2. Perform an operation that touches encryption on it — for example creating a volume from it.
3. Confirm the driver reports an error and keeps running, rather than the controller pod restarting.

Automated coverage is included: the new test reproduces the crash against the unfixed code.

### Why make this change?

The check ended by reading a value that was not always set. For a filesystem-backed volume with no API client, the driver has no way to ask the cluster whether the volume is encrypted, so nothing filled that value in and reading it crashed the process.

Reporting an error is the safe answer rather than assuming "not encrypted". Callers use this to decide whether to apply encryption, so a volume whose state could not be read would otherwise be treated as one that had been read and found unencrypted.

Found by @kristina-solovyova in #692.

### DIFF
--- a/pkg/wekafs/volume.go
+++ b/pkg/wekafs/volume.go
@@ -189,7 +189,8 @@ func (v *Volume) isEncrypted(ctx context.Context) (bool, error) {
 
 	if v.encrypted == nil {
 		// TODO: maybe we should move it into a separate method that is explicitly invoked when Volume is constructed from VolumeID rather than doing it here
-		if exists, err := v.Exists(ctx); err == nil {
+		exists, existsErr := v.Exists(ctx)
+		if existsErr == nil {
 			if v.apiClient != nil {
 				fsObj, err := v.getFilesystemObj(ctx, true)
 				if err != nil {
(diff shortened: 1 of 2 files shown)


sergeyberezansky:
* **#793** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/793?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/793?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

PR #794, opened 2026-08-13
docs: show a Source Code link on the chart page

sergeyberezansky:
### TL;DR

The chart page now shows a **Source Code** link, pointing at the tag the chart was built from.

### What changed?

- The chart README and the ArtifactHub listing render a `## Source Code` section, taken from the `sources` entry already in `Chart.yaml`.
- Fixed the source URL recorded by PR-built charts, which contained the literal text `$CHART_VERSION` instead of the version number. Released charts were not affected.

### How to test?

1. Open `charts/csi-wekafsplugin/README.md` and confirm a **Source Code** section appears under **Maintainers**, with a link ending in the chart version.
2. On a PR build, run `helm show chart <chart>` and confirm the `sources` URL contains a version number rather than `$CHART_VERSION`.

### Why make this change?

`Chart.yaml` has always recorded where the chart comes from, but nothing displayed it, so the published chart page gave no link back to the source at that version. Rendering it from `Chart.yaml` means it follows the value the release process already maintains, rather than becoming another thing to update by hand.

Both sections were originally raised by @ari in #582.

### DIFF
--- a/.github/workflows/pull-request-v2.yaml
+++ b/.github/workflows/pull-request-v2.yaml
@@ -333,7 +333,7 @@ jobs:
             APP_VERSION="${{ steps.version.outputs.version }}"
             yq -i '.version = "'$CHART_VERSION'"' $BASEDIR/Chart.yaml
             yq -i '.appVersion = "'$APP_VERSION'"' $BASEDIR/Chart.yaml
-            yq -i '.sources[0] = "https://github.com/weka/csi-wekafs/tree/v$CHART_VERSION/charts/csi-wekafsplugin"' $BASEDIR/Chart.yaml
+            yq -i '.sources[0] = "https://github.com/weka/csi-wekafs/tree/v'$CHART_VERSION'/charts/csi-wekafsplugin"' $BASEDIR/Chart.yaml
             yq -i '.csiDriverVersion = "'$DRIVER_VERSION'"' $BASEDIR/values.yaml
             echo ------------------ values ------------------
             cat $BASEDIR/values.yaml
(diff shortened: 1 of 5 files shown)


sergeyberezansky:
* **#794** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/794?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/794?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

sergeyberezansky:
### Why the CI change is in this PR

Rendering the section is what exposed the value behind it. `pull-request-v2.yaml` wrote `sources[0]` with the variable inside single quotes, so it never expanded:

```bash
# pull-request-v2.yaml (both occurrences) — $CHART_VERSION stays literal
yq -i '.sources[0] = ".../tree/v$CHART_VERSION/charts/csi-wekafsplugin"'

# push-dev-v2.yaml — breaks out of the quotes, expands correctly
yq -i '.sources[0] = ".../tree/v'$CHART_VERSION'/charts/csi-wekafsplugin"'
```

Verified in a shell: the first form yields `.../tree/v$CHART_VERSION/...`, the second `.../tree/v2.9.2/...`. Both occurrences now match the form the other three workflows already use. `release-v2.yaml` was always correct, so released charts never carried the broken URL — only PR builds did, and nothing rendered it until now.

### On the Maintainers half of #582

Not needed. `chart.maintainersSection` emits its own `## Maintainers` heading, so the gotmpl was already complete and both READMEs already render it. #582 added a literal heading above the template, which would have produced the heading twice. Confirmed against the generated output rather than the template — the section count is 1 in each README after this change.

### On the rest of #582

`docs/NFS.md`, `docs/deployment.md`, `docs/storage-class-configurations.md` and `migration/upgrade-legacy-pv.md` are byte-identical to #784, so they are already carried there. `selinux/README.md` was checked and deliberately not taken: dev's version is the newer one documenting CIL-only installation, while #582's documents the pre-CIL `checkmodule`/`.pp` flow referencing files that no longer ship. That closes out everything from #582.

PR #818, opened 2026-09-14
fix: make "-option" mount overrides actually remove the option

sergeyberezansky:
### TL;DR
`-sync_on_close` in a PVC or pod annotation now actually removes the option, and a Weka version that cannot accept an option no longer gets it anyway.

### What changed?
- A `-option` mount override is recorded as an exclusion, so it survives the driver defaults being merged in afterwards; a later `+option` still puts it back
- `sync_on_close` is now dropped from the options actually mounted with on clusters below Weka v4.2, instead of being pruned and then re-added by the defaults
- `ro` and `container_name` are still refused when supplied as custom mount options, and are now also refused when supplied through an annotation override
- The documented order of application was wrong about when the defaults apply; corrected

### How to test?
1. Set `allowMountOptionOverrides: true` and annotate a PVC with `weka.io/mount-options-override: -sync_on_close`
2. Create a pod using that PVC
3. On the node, run `mount | grep wekafs` — the options should no longer contain `sync_on_close`
4. Change the annotation to `-sync_on_close,+sync_on_close` and remount — it should be present again
5. Attach a volume read-only and check the mount still carries `ro`

### Why make this change?
Removing an option only worked if it came from the StorageClass. Anything the driver adds by default — `sync_on_close`, `writecache` — was silently put back after the override ran, so the annotation appeared to be ignored with nothing in the log to explain it. The same gap meant a cluster too old for `sync_on_close` was still mounted with it.

---

> [!NOTE]
> **Medium Risk**
> Changes what mount flags actually reach the kernel for published volumes, including readonly and cluster-version gating, though logic is heavily tested and scoped to option assembly.
> 
> **Overview**
> **`-option` overrides (PVC/pod annotations) now stick** by recording removals in `excludeOptions` via new `ExcludeOption` / `UnexcludeOption`, instead of only deleting from the volume’s option set. That matters because node defaults are merged **under** volume options at `MountUnderlyingFS`, which previously reintroduced options like `sync_on_close` and `writecache` after an override.
> 
> Override parsing uses **`ExcludeOption` for `-opt`** and **`UnexcludeOption` before adds** so `-opt,+opt` and valued options behave correctly; readonly attachment uses `ExcludeOption("rw")` when adding `ro`.
> 
> **Pruning is split and re-timed:** cluster-capability drops (e.g. `sync_on_close` on old Weka) run on the **fully merged** options at mount time; user-forbidden options (`ro`, `container_name`) are stripped only on user/annotation layers, with an extra prune in `NodePublishVolume` after overrides so annotations cannot bypass those rules.
> 
> Documentation in `constants.go` updates the mount-option application order; broad unit tests cover exclusion, defaults merge, and the full cache-option pipeline.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2833fec02cf2264f0fd1b8309104b8fb32a83e52. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sergeyberezansky:
* **#818** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/818?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/818?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

sergeyberezansky:
### Merge activity

* **Sep 15, 11:13 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/weka/csi-wekafs/818).
* **Sep 15, 11:14 AM UTC**: @sergeyberezansky merged this pull request with [Graphite](https://app.graphite.com/github/pr/weka/csi-wekafs/818).

PR #819, opened 2026-09-15
test: cover the mount option pipeline as NodePublishVolume actually runs it

sergeyberezansky:
### TL;DR

Test-only. Nothing changes for anyone running the driver.

### What changed?

- A test that assembles mount options the way `NodePublishVolume` really does, so the steps that reject options a user is not allowed to set are covered for the first time.
- A test pinning `+option` and bare `option` as equivalent in an override annotation.
- No production code touched.

### How to test?

Automated only: `go test ./pkg/wekafs/ -run 'MountOption'`.

### Why make this change?

The previous test covered only how options merge, so it never reached the checks that drop options an older Weka cluster cannot take, or refuse `ro` and `container_name` from user-supplied options. A removal that silently failed, or an override that smuggled in a refused option, would not have been caught.

---

> [!NOTE]
> **Low Risk**
> Test-only changes with no production code or runtime behavior modified.
> 
> **Overview**
> Adds integration-style tests in `mountoptoverrides_test.go` so mount-option behavior matches **`NodePublishVolume`**, not only the lighter `mountOptionPipeline` helper.
> 
> **`TestMountOptionPipeline_FullPublishScenario`** wires real `Volume` / `NodeServer` steps: storage class options, `pruneUnsupportedMountOptions`, PVC/pod annotation overrides via `applyMountOptionsOverridesToVolume`, read-only merge, and the final defaults merge. It locks in the reported user case (`writecache` + `-sync_on_close,ro,readcache` → **`readcache` only**, with user **`ro` refused** and **`sync_on_close` excluded**), across bare vs `+`-prefixed overrides and PVC-only vs PVC+pod splits. Control subtests separate driver-added **`ro`**, exclusion vs cluster capability pruning (fake `apiclient` with `SyncOnCloseMountOption: true`), and nil-client capability drops.
> 
> **`TestMountOptionOverride_BarePrefixEquivalence`** asserts bare and `+`-prefixed modifiers produce identical `MountOptions` (including `excludeOptions`), across valued options, mixed removals, re-adds, exclusives, and whitespace.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb6c7ee7b213d813510a68dadfdb163cd435edcc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sergeyberezansky:
* **#820** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/820?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#819** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/819?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/819?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

sergeyberezansky:
Verification detail, kept out of the description since it is developer-facing.

### What the new test actually exercises

The pre-existing `mountOptionPipeline` helper models only `Merge`, so `pruneUnsupportedMountOptions` was unreachable from any test. `TestMountOptionPipeline_FullPublishScenario` drives the real methods in the order `NodePublishVolume` uses them — `setMountOptions` → prune → capability-flag merge → the real `applyMountOptionsOverridesToVolume` against PVC and Pod objects carrying real annotations → prune → readonly merge → the defaults merge in `MountUnderlyingFS`. Only the assembly glue is new; nothing is reimplemented.

Measured result for the case that prompted this — StorageClass `writecache`, node defaults adding `sync_on_close`, override `-sync_on_close,ro,readcache`:

```
final mount options: "readcache"
```

`readcache` displaces `writecache`, the exclusion removes `sync_on_close`, and `ro` is refused. Identical across all four forms: bare vs fully `+`-prefixed, and the override whole on the PVC vs split across PVC and Pod.

### Controls

Two subtests exist to stop the main assertions passing for the wrong reason.

- **`ro` refusal vs the driver's own `ro`** — a readonly attachment with no override yields `ro,sync_on_close,writecache`, confirming that refusing a user-supplied `ro` has not broken readonly volumes. The refusal runs before the driver adds its own.
- **Exclusion vs capability pruning** — both remove `sync_on_close`, so a nil `apiClient` would hide which one did. Using `&apiclient.ApiClient{CompatibilityMap: &apiclient.WekaCompatibilityMap{SyncOnCloseMountOption: true}}` (no cluster needed), `-sync_on_close` alone gives `writecache`, while the same client with no override keeps `sync_on_close,writecache`. So the exclusion is demonstrably what survives the defaults merge.

### Mutation check on the equivalence test

`TestMountOptionOverride_BarePrefixEquivalence` compares whole `MountOptions` values, including `excludeOptions`, rather than asserting presence. To confirm it can actually fail, the `+` branch of `ApplyToOptions` was temporarily rewritten to skip `UnexcludeOption`:

```
--- FAIL: TestMountOptionOverride_BarePrefixEquivalence/re-add_after_removal_(UnexcludeOption_path)
```

Production code was restored immediately; the diff on this branch is the single test file.

### Not addressed here

`NodePublishVolumeError` and `NodeUnpublishVolumeError` still apply `strings.ToLower` to the message, which flattens case-sensitive filesystem and snapshot names in anything an operator reads. Left alone deliberately — the same applies to the six controller helpers, so it belongs in its own change.

sergeyberezansky:
### Merge activity

* **Sep 15, 12:43 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/weka/csi-wekafs/819).
* **Sep 15, 12:43 PM UTC**: @sergeyberezansky merged this pull request with [Graphite](https://app.graphite.com/github/pr/weka/csi-wekafs/819).

assafgi (approved):

PR #825, opened 2026-09-15
fix: keep a readonly attachment readonly when an override removes "ro"

sergeyberezansky:
### TL;DR

A volume attached read-only stayed read-only for the pod, but the driver's own mount of the underlying Weka filesystem was made writable if a PVC or pod annotation asked to remove `ro`.

### What changed?

- `-ro` in a `weka.io/mount-options-override` or `weka.io/mount-options-overrides` annotation no longer affects the `ro` that a read-only volume attachment applies.
- Regression tests for a read-only attachment combined with a `-ro` override, on the PVC and on the pod, plus the mirror case confirming a `-ro` override still adds no `ro` when the attachment is not read-only.

### How to test?

1. Deploy the driver with `mountOptionOverrides.enabled=true`.
2. Create a PVC with the annotation `weka.io/mount-options-override: "-ro"`.
3. Mount it in a pod through a read-only volume (`readOnly: true` on the pod's `persistentVolumeClaim`).
4. On the node, run `mount -t wekafs`. The Weka mount for that volume should show `ro`. Before this change it showed `rw`.

Also covered by `go test ./pkg/wekafs/ -run FullPublishScenario`.

### Why make this change?

`ro` is not accepted as a user-supplied mount option — a read-only attachment is the supported way to ask for it — so the driver refuses it from annotations. The refusal removed the option but not the instruction to remove it, and that instruction was applied after the driver had added its own `ro`, cancelling it. The pod was never writable, since the mount into the container carries its own `ro` independently, but the driver was no longer doing what it reported, and would have started mounting writable in any future path that relies on the volume's own options.

---

> [!NOTE]
> **Medium Risk**
> Changes mount-option merge behavior on publish for volumes using `-ro` overrides; wrong handling could affect read-only vs read-write on the driver’s parent mount, though scope is limited to that override pattern.
> 
> **Overview**
> Fixes a bug where a **readonly** volume attachment could end up with a **writable** parent Weka mount when PVC/pod overrides included `-ro`.
> 
> `withRejectedCustomMountOptionsPruned` already strips user-supplied `ro`/`+ro` (readonly must come from the attachment), but `-ro` only recorded an **exclusion** that survived that step. The driver still adds `ro` for readonly attachments, yet at the node-defaults merge exclusions run after additions—so the stale exclusion removed `ro` and the underlying FS mounted read-write.
> 
> The refusal path now also calls **`UnexcludeOption(MountOptionReadOnly)`**, treating `-ro` like other disallowed `ro` input without leaving exclusion bookkeeping behind. Clearing the exclusion does **not** add `ro` when the attachment is not readonly.
> 
> **Tests** extend the full `NodePublishVolume` mount-option pipeline: readonly + `-ro` (PVC, pod, mixed modifiers) must keep `ro`; writable attachment + `-ro` must not gain `ro`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c7ac45dab1815e68bc3da0edde3ce5165f2f3c88. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sergeyberezansky:
* **#825** <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/825?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/weka/csi-wekafs/825?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

